PR #2, opened 2017-12-08
Implement bash/zsh completion framework

thetechnick:
- Added new `hcloud completion` command
- `hcloud server delete/describe <TAB>` will now autocomplete server ids
- all commands are autocompleted
- added `--no-header` flag to `hcloud server list` for the autocomplete code

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -1,0 +1,220 @@
+package cli
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	completionShells = map[string]func(out io.Writer, cmd *cobra.Command) error{
+		"bash": runCompletionBash,
+		"zsh":  runCompletionZsh,
+	}
+	completionShortDescription = "Output shell completion code for the specified shell (bash or zsh)"
+	completionLongDescription  = completionShortDescription + `
+
+Note: this requires the bash-completion framework, which is not installed by default on Mac. This can be installed by using homebrew:
+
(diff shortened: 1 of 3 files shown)


server_list.go
```diff
@@ -16,19 +16,23 @@ func newServerListCommand(cli *CLI) *cobra.Command {
 		TraverseChildren: true,
 		RunE:             cli.wrap(runServerList),
 	}
+	cmd.Flags().Bool("no-header", false, "Do not print table header")
```
thcyron: Please do not include these changes in this PR.

Let’s discuss separately how we want to allow the user of the CLI to configure how the output should look like (table with headers, table without headers, just selected columns, JSON, YAML, etc.).

root.go
```diff
@@ -2,15 +2,38 @@ package cli
 
 import "github.com/spf13/cobra"
 
+const (
+	bashCompletionFunc = `
```
thcyron: What about moving this completion-related constant to `completion.go`?

completion.go
```diff
@@ -0,0 +1,220 @@
...
+		Long:      completionLongDescription,
+		RunE:      cli.wrap(runCompletionCommand),
+		ValidArgs: shells,
+	}
+	return cmd
+}
+
+func runCompletionCommand(cli *CLI, cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return errors.New("shell not specified")
```
thcyron: What about doing validation in a `PreRunE` function like already done in `validateSSHKeyCreate()`?

PR #26, opened 2018-01-12
Add server change-type command

thetechnick:
Closes #26

### DIFF
--- a/server_change_type.go
+++ b/server_change_type.go
@@ -1,0 +1,63 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/hetznercloud/hcloud-go/hcloud"
+	"github.com/spf13/cobra"
+)
+
+func newServerChangeTypeCommand(cli *CLI) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "change-type [FLAGS] SERVER SERVERTYPE",
+		Short:                 "Change the server type of a server",
+		Args:                  cobra.ExactArgs(2),
+		TraverseChildren:      true,
+		DisableFlagsInUseLine: true,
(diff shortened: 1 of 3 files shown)


server_change_type.go
```diff
@@ -0,0 +1,64 @@
...
+	if serverType == nil {
+		return fmt.Errorf("server type not found: %s", serverTypeIDOrName)
+	}
+
+	keepDisk, _ := cmd.Flags().GetBool("keep-disk")
+	opts := hcloud.ServerChangeTypeOpts{
+		ServerType:  serverType,
+		UpgradeDisk: !keepDisk,
+	}
+	fmt.Printf("%#v", opts)
```
thcyron: 🐒

server_change_type.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := cli.ActionProgress(cli.Context, action); err != nil {
+		return err
+	}
+
+	if opts.UpgradeDisk {
+		fmt.Printf("server %s changed to type %s\n", idOrName, serverTypeIDOrName)
```
thcyron: Should start with an uppercase letter to be consistent with other commands.

server_change_type.go
```diff
@@ -0,0 +1,64 @@
...
+	} else {
+		fmt.Printf("server %s changed to type %s disk size was unchanged\n", idOrName, serverTypeIDOrName)
```
thcyron: Would prefer to put “disk size was unchanged” in parenthesis:

```
"Server %s changed to type %s (disk size was unchanged)\n"
```

PR #28, opened 2018-01-12
Add image update command

### LINKED ISSUE #8: Implement image update command
```
hcloud image update ID --description DESCRIPTION --type TYPE
```

thetechnick:
Closes #8

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -68,7 +68,7 @@ const (
 	}
 
 	__hcloud_image_types_no_system() {
-		COMPREPLY="snapshot backup"
+		COMPREPLY=($(echo -e "snapshot\nbackup"))
 	}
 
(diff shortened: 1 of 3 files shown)


completion.go
```diff
@@ -68,7 +68,7 @@ const (
 	}
 
 	__hcloud_image_types_no_system() {
-		COMPREPLY="snapshot backup"
```
thcyron: I think your previous code would work, but you missed the parens: `COMPREPLY=("snapshot" "backup")`

thcyron:
I leave it to you if you want to change `COMPREPLY`.

completion.go
thetechnick: does not work :(

PR #71, opened 2018-01-17
Set cli version on compile

thetechnick:
Build release using:
```shell
./script/build.bash linux amd64 `./script/git-version.bash`
```

### DIFF
--- a/version.go
+++ b/version.go
@@ -6,7 +6,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const Version = "1.0.0"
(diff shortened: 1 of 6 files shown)


version.go
```diff
@@ -6,7 +6,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const Version = "1.0.0"
+// Version is set via compiler flags (see script/build.bash)
+var Version = "was not build properly"
```
thcyron: buil**t**

PR #111, opened 2018-05-23
Add command to spawn an SSH connection to the server

### LINKED ISSUE #110: Consider 'hcloud server ssh NAME' command
It would be very handy to have a command

    hcloud server ssh NAME

to spawn an `ssh` command connecting to the given server.

Opinions?

Nikola-K:
Closes #110

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## v1.5.0
+
+* Add `ssh` command for servers to start a new SSH connection to the server
+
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+## v1.5.0
```
thcyron: Use `## master`, the exact version is replaced when it is actually released.

PR #116, opened 2018-07-30
Print IP address after a server was created

### LINKED ISSUE #107: Show IP address after server has been created
Print the server’s IP address after the server has successfully been created.

philippgille:
The previous output was one line with one (ID) or two (ID and root password) info,
so instead of adding another one to the line, or having two lines with potentially
different amount of info, I also changed the printed output after creating a server.
It's now one short line per info: ID, IP address, optional root password.

Example:

```bash
./hcloud server create --image "ubuntu-16.04" --name "test" --type "cx11"
  10s [====================================================================] 100%
Server created with ID: 920920
Root password: secret
IP address: 123.123.123.123
```

Let me know if you have any improvement suggestions

Closes #107

### DIFF
--- a/cli/server_create.go
+++ b/cli/server_create.go
@@ -71,10 +71,15 @@ func runServerCreate(cli *CLI, cmd *cobra.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


thcyron:
Thank you!

Issue #117, opened 2018-08-06
Action progress bar: Add option for disabling the progress bar


LKDevelopment:
When you create a large number of servers, with the watching of an action progress you could run really fast into the API limits. It would be nice if there would be an option to easily disable the progress bar.

thcyron:
Done (sort of) with #119.

PR #118, opened 2018-08-06
Add Flag for easily disable the progress pulling from the api.

LKDevelopment:
This pr aims to solve a possible issue for people that could run in the API limits when they create many servers througth the CLI. It adds a boolean flag to the server creation command called "disable-progress". Default is set to "false" so it is the same behavior as in the moment. 

Issue for this: #117


cli/server_create.go
```diff
@@ -67,8 +69,11 @@ func runServerCreate(cli *CLI, cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := cli.ActionProgress(cli.Context, result.Action); err != nil {
-		return err
+	disableProgressBar, _ := cmd.Flags().GetBool("disable-progress")
+	if disableProgressBar == false {
```
thcyron: You don’t only disable progress reporting; you disable action error handling altogether.

cli/cli.go
```diff
@@ -116,6 +117,10 @@ func (c *CLI) Client() *hcloud.Client {
 		if c.Endpoint != "" {
 			opts = append(opts, hcloud.WithEndpoint(c.Endpoint))
 		}
+		pollInterval, _ := c.RootCommand.PersistentFlags().GetInt("poll-interval")
```
thcyron: There’s a [`GetDuration()`](https://godoc.org/github.com/spf13/pflag#FlagSet.GetDuration) you can use. Allows to specify the interval with suffixes like 1s, 500ms, etc.

cli/root.go
```diff
@@ -1,6 +1,9 @@
 package cli
 
-import "github.com/spf13/cobra"
+import (
+	"github.com/spf13/cobra"
+	"time"
+)
```
thcyron: Please format the source files using [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports).

PR #138, opened 2018-09-29
Add `start-after-create` flag to server create cmd

artheus:
Reading the [hcloud documentation](https://docs.hetzner.cloud/#servers-create-a-server) I found out that there is a Boolean flag, that one can send, which enables you to choose if the server should start directly after creation.

This is a good feature to include if one wants to install alternate operating systems from the default vm images provided by Hetzner Cloud.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## master
+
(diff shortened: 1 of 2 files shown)


thcyron:
Thank you for your contribution!

Issue #146, opened 2018-11-13
Remove --window parameter from enable-backup command


thcyron:
As specifying a backup window is no longer supported, that parameter is no longer needed.

PR #147, opened 2018-11-13
Deprecate `--window` flag on `hcloud server enable-backup`

LKaemmerling:
#146


cli/server_enable_backup.go
```diff
@@ -19,9 +19,6 @@ func newServerEnableBackupCommand(cli *CLI) *cobra.Command {
 	cmd.Flags().String(
 		"window", "",
 		"The time window for the daily backup to run. All times are in UTC. 22-02 means that the backup will be started between 10 PM and 2 AM.")
```
thcyron: What about adding a deprecation notice to the flag’s usage description?

LKaemmerling: I added a `(deprecated)` to the beginning of the sentence, because so you see it directly.

PR #149, opened 2018-11-15
Wait Next Actions

LKaemmerling:
Update hcloud-go to v1.11.0
Wait for the new next actions on server create and volume create

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -176,7 +176,7 @@ func (c *CLI) WaitForActions(ctx context.Context, actions []*hcloud.Action) erro
 
 		switch action.Command {
 		default:
-			fmt.Print("Waiting for action to have finished... ")
+			fmt.Printf("Waiting for action %s to have finished... ", action.Command)
 		case "start_server":
 			fmt.Printf("Waiting for server %d to have started... ", resources["server"])
 		case "attach_volume":
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -167,16 +167,29 @@ func (c *CLI) ensureToken(cmd *cobra.Command, args []string) error {
 }
 
 func (c *CLI) WaitForActions(ctx context.Context, actions []*hcloud.Action) error {
-	if len(actions) > 0 {
-		for _, action := range actions {
-			fmt.Print(actionToString(action))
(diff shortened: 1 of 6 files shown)


cli/util.go
```diff
@@ -86,3 +88,16 @@ func labelsToString(labels map[string]string) string {
 	}
 	return strings.Join(labelsString, ", ")
 }
+
+func actionToString(action *hcloud.Action) string {
+	var actionString string
+	switch action.Command {
+	default:
+		actionString = "Waiting for actions to have finished"
```
thcyron: Should be singular.

cli/util.go
```diff
@@ -86,3 +88,16 @@ func labelsToString(labels map[string]string) string {
...
+	case "start_server":
+		actionString = fmt.Sprintf("Waiting for server %d to have started", action.Resources[0].ID)
+	case "attach_volume":
+		actionString = fmt.Sprintf("Waiting for volume %d to have been attached to server %d", action.Resources[0].ID, action.Resources[1].ID)
```
thcyron: I don’t think you should rely on the ordering of `action.Resources`.

cli/util.go
```diff
@@ -86,3 +88,16 @@ func labelsToString(labels map[string]string) string {
 	}
 	return strings.Join(labelsString, ", ")
 }
+
+func actionToString(action *hcloud.Action) string {
```
thcyron: Having `actionToString` return a string of the form “Waiting for …” does not make sense to me; especially not when considering how `labelsToString` and the other similar functions work.

I would inline the `switch` to `WaitForActions`.

cli/cli.go
```diff
@@ -164,3 +165,18 @@ func (c *CLI) ensureToken(cmd *cobra.Command, args []string) error {
 	}
 	return nil
 }
+
+func (c *CLI) WaitForActions(ctx context.Context, actions []*hcloud.Action) error {
+	if len(actions) > 0 {
```
thcyron: FYI: Not needed since you can range over empty/nil slices.

cli/util.go
thcyron: Maybe add the command to the message?

Issue #158, opened 2019-01-17
Allow command to be send to server through hcloud ssh


joto:
`hcloud server ssh` allows me to login to a server, but I can't send a command to be run on the server like with a normal `ssh some-command`. It would be nice to have this to easily run commands on the server.

PR #160, opened 2019-01-27
Allow commands to be send through "hcloud ssh".

joto:
Implementation for #158.

### DIFF
--- a/cli/server_ssh.go
+++ b/cli/server_ssh.go
@@ -12,9 +12,9 @@ import (
 
 func newServerSSHCommand(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                   "ssh [FLAGS] SERVER",
+		Use:                   "ssh [FLAGS] SERVER [COMMAND...]",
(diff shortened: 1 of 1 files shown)


thcyron:
Thanks for your contribution!

I noticed that executing commands with flags is a bit cumbersome as you need to use a `--` to separate the flags for the command to execute from the flags for the cli itself:

```
$ ./hcloud server ssh testcli uname -a
hcloud: unknown shorthand flag: 'a' in -a
$ ./hcloud server ssh testcli -- uname -a
Linux testcli 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux
```

Could you please have a quick look if there is a way to make `./hcloud server ssh testcli uname -a` actually run `uname -a`?

thcyron:
`flags.SetInterspersed(false)` is what we’re looking for (as seen in the [docker cli source code](https://github.com/docker/cli/blob/master/cli/command/container/exec.go)).

Issue #154, opened 2018-12-14
Allow to config the context with a file in the current path


c33s:
as i have multiple customers on hetzer servers which all have their own account i would love to have a file like `.hcloud` where i can specify the current context. so switching between directories will automatically switch context for hcloud (never work in production context while thinking to work in the development context)

having the following directory tree:
````
project dir
├───acme
│       .hcloud (contains: `context: acme_production`)
│
├───acme_development
│       .hcloud (contains: `context: acme_development`)
│
├───foobar
│       .hcloud (contains: 'context: foobar_production`)
│
└───foobar_development
        .hcloud (contains: `context: foobar_development`)
```
so swtiching to the directory `acme` and calling `hcloud context active` will show `acme_production`, switching to the directory `foobar_development` running `hcloud context active` there will show `foobar_development`.

thcyron:
I implemented that in the `autocontext` branch: https://github.com/hetznercloud/cli/tree/autocontext

Just put the name of the context to activate in a file called `.hcloud-context`.

Please give it a shot.

c33s:
@thcyron 
> Please give it a shot.

would love to try but it looks a little tricky to compile a specific branch. any hints?

c33s:
thanks for the build support. works like charm.

aaaawwwsome! the simple config file implementation also works if called from my  [cmder](http://cmder.net/) lua script which i use for my command prompt. 

![2019-01-29 15_13_11-cmder](https://user-images.githubusercontent.com/649209/51914411-1349f480-23d9-11e9-84f1-6d94d9618375.png)

gitbranch | hetzner cloud context | terraform workspace

divramod:
the branch you mentioned is gone, is it still somewhere around here? if yes, where can i find it? thx in advance!

c33s:
to dig out the discussion/decision from the PR https://github.com/hetznercloud/cli/pull/162:

@thcyron:

> After discussing this PR in our team, we decided to not continue with this particular implementation with a .hcloud-context file. Until another proposal is submitted, you would need to utilize the already existing HCLOUD_* environment variables.

https://github.com/hetznercloud/cli/pull/162#issuecomment-460320408

which was from my point of view a bad decision leaving me with a self compiled old version of hcloud cli. many people don't understand that environment variables are not the holy grail. they are awesome but not the solution for everything. config files are still comfortable und useful. keeping it like this is a bad DX (developer experience)

@c33s:

[...]

PR #162, opened 2019-01-29
Load context from .hcloud-context file

thcyron:
When `hcloud` is invoked and there is a file called `.hcloud-context` in the current working directory that contains the name of a context, that context is automatically activated.

However, similar functionality can be achieved by using third-party tools like [direnv](https://direnv.net) which set per-directory environment variables. By configuring the `HCLOUD_CONTEXT` variable, you basically end up with the same effect as `.hcloud-context`. As a bonus, you could also set `HCLOUD_CONFIG` and `HCLOUD_TOKEN`.

Features has been requested in #154.

## TODO

* [ ] Add changelog line

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,10 +1,12 @@
 package cli
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -50,6 +52,35 @@ func (c *CLI) ReadEnv() {
 	}
 	if s := os.Getenv("HCLOUD_CONTEXT"); s != "" && c.Config != nil {
 		c.Config.ActiveContext = c.Config.ContextByName(s)
+	}
+}
+
+const ContextFileName = ".hcloud-context"
+
+func (c *CLI) ReadContextFile() {
+	if c.Config == nil {
+		return
+	}
+
+	data, err := ioutil.ReadFile(ContextFileName)
+	if os.IsNotExist(err) {
+		return
+	}
+	if err != nil {
+		log.Printf("warning: failed to read %s file: %s\n", ContextFileName, err)
+		return
+	}
+
+	name := string(bytes.TrimSpace(data))
+	if name == "" {
+		log.Printf("warning: ignoring invalid context in %s\n", ContextFileName)
+		return
+	}
+	if context := c.Config.ContextByName(name); context != nil {
+		c.Config.ActiveContext = context
+		c.Token = context.Token
+	} else {
+		log.Printf("warning: context %q specified in %s does not exist\n", name, ContextFileName)
 	}
 }
 
--- a/cmd/hcloud/main.go
+++ b/cmd/hcloud/main.go
@@ -30,6 +30,7 @@ func main() {
 		}
 	}
 
+	c.ReadContextFile()
 	c.ReadEnv()
 
 	if err := c.RootCommand.Execute(); err != nil {


cli/cli.go
```diff
@@ -53,6 +55,35 @@ func (c *CLI) ReadEnv() {
 	}
 }
 
+const ContextFileName = ".hcloud-context"
```
c33s: shouldn't the file be named `.hcloud`. in the future there maybe will be more directory specific configuration beside the context.
```suggestion
const ContextFileName = ".hcloud"
```

thcyron: Valid point. However, I would prefer to make using that feature as easy as possible and not require the user to write TOML/YAML/JSON files:

    echo my-context > .hcloud-context

If you need per-directory configs, maybe you should use tools like [direnv](https://direnv.net) which allows you to set per-directory environment variables. By setting `HCLOUD_CONFIG` you can configure `hcloud` to use a different config file per directory:

```
$ cat .envrc
export HCLOUD_CONFIG=.hcloud.toml
```

Actually, you could even use direnv to replace this pull request:

```
$ cat .envrc
export HCLOUD_CONTEXT=my-context
```

To be honest, I tend to prefer the generic direnv solution over our custom implementation.

c33s: > To be honest, I tend to prefer the generic direnv solution over our custom implementation.

this would be again, another dependency. the stack grows and grows (vagrant, puppet, terraform, hcloud). i vote strongly against using direnv.
your solution simply works. it has minimal performance impact but if a file is there it is read.

symfony (php framework) made the same step, first it removed `parameters.yaml` to move to `.env` but `.env` was only used for dev not for production. then the move "back". refactor of the `.env` component and also the official statement, that it can be used in production.

having this config file hurts nobody and gives another option without adding another tool to the toolchain but it don't prevent you from using direnv.

we are software architects, the big advantage over building architects is, that we can have both a "flat top" and a "pitched roof" at the same time.

c33s: just had a look at direnv, it doesn't look like to have full windows support. if i don't use bash i can't do a `eval "$(direnv hook bash)"` so it won't run as transparent as a simple config file.
often i cannot easily control which shell a process spans. 

- will terraform span a bash shell where it will load my profile where the direnv is loaded?
- will the direnv smoothly work within a bash shell called with terraform on windows?
- how will hooking behave if called from terrafrom?

sounds like a lot of overhead, config and testing for simple context switching.

having a full `TOML/YAML/JSON` file with a full possibilities for configuring hcloud would be awesome. it makes a really good developer expierence without killing performance. the only performance overhead is one `if file_exists()` check. no pain and much gain :)

cli/cli.go
```diff
@@ -53,6 +55,35 @@ func (c *CLI) ReadEnv() {
...
+	}
+
+	name := string(bytes.TrimSpace(data))
+	if name == "" {
+		log.Printf("warning: ignoring invalid context in %s\n", ContextFileName)
+		return
+	}
+	if context := c.Config.ContextByName(name); context != nil {
+		c.Config.ActiveContext = context
+		c.Token = context.Token
```
c33s: some output would be good, showing something like `temporary switching to context your-context`. if called from terraform or any automation script it would be helpful to have such output in the logs. could be hard to debug being in an auto-context and not knowing it.

also some kind of control for this output would be good maybe a `--verbose` option to show it or `--quiet` to supress it.

the output is currently like that if called from a lua script (i call it twice) for showing the context in the command prompt:
```
hcloud: warning: context "not-existing-context" specified in .hcloud-context does not exist
hcloud: warning: context "not-existing-context" specified in .hcloud-context does not exist
C:\terraform\project (development) (hcc: c33s-development)(tfws: development)
```
so being able to call it with `--quiet` from my lua script would be cool.

thcyron:
After discussing this PR in our team, we decided to not continue with this particular implementation with a `.hcloud-context` file. Until another proposal is submitted, you would need to utilize the already existing `HCLOUD_*` environment variables.

Nevertheless, thank you for your feedback, @c33s.

Issue #82, opened 2018-01-29
Structured access to fields (JSON, Go templates)


thcyron:
Sometimes you want to use the CLI in a shell script to get some information, but parsing the output is obviously a bad idea. We should consider adding JSON support so you can access information in a structured way.

Have a look at how `docker inspect` enables this: https://docs.docker.com/engine/reference/commandline/inspect/#extended-description

hojerst:
isn‘t this already possible by using

hcloud -o outputformat=json -o columns=field1,field2

?

thetechnick:
@hojerst This is not yet implemented `-o` only supports `noheader` and `columns=` but not json output or templating.

We are still discussing how we want to implement this functions and what we want to support.

mblaschke:
Any news about this feature?
Parsing strings doesn't feel good. Would be nice to get a valid JSON structure which can be parsed.

thcyron:
I’d suggest you use [our API](https://docs.hetzner.cloud/) directly if you need JSON output.

mblaschke:
So i have to write my own cli client to get the json output?

I like the `docker` and `kubectl` approach where you can get the json output directly and also the readable formats. Should be not too difficult to implement this feature.

This feature would allow easy and safe processing in other scripting languages and just use the hcloud cli client to access the api.

thcyron:
There’s now support for JSON and Go template output. See the readme how to use it.

PR #164, opened 2019-01-29
Add -g/--get option to "server describe" command.

joto:
This allows getting parameters of a server in a way that's easy to use in scripts. The normal "server describe" command outputs the data in a human readable form, but in (shell) scripts it would be much easier to just have single parameters printed.

This pull request is preliminary, it only implements the `public_net.ipv4.ip` parameter, but I can extend this, if you think this is going in the right direction.

An alternative would be to implement this as a new command, something like "server get" or so.

### DIFF
--- a/cli/server_describe.go
+++ b/cli/server_describe.go
@@ -17,6 +17,7 @@ func newServerDescribeCommand(cli *CLI) *cobra.Command {
 		PreRunE:               cli.ensureToken,
 		RunE:                  cli.wrap(runServerDescribe),
 	}
+	cmd.Flags().StringP("get", "g", "", "Get parameter")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


thcyron:
Basically you’re asking for #82. I’d prefer a generic solution for all resources over a specific solution just for servers and a few attributes.

joto:
Well, we can extend this to all attributes and to more commands. But it is a lot of busy code, I agree. I had thought about a more generic solution writing out the JSON received from the server, but that doesn't seem to be easily available in a convenient place.

thcyron:
Printing the JSON is actually quite easy.

https://github.com/hetznercloud/cli/blob/master/cli/server_describe.go#L25

The second return value is a `*http.Response` whose `Body` is the JSON data. We could print that (probably pretty-printed) and by piping it into `jq` you could easily access specific fields:

```
$ hcloud server describe 1234 --json | jq -r '.server.public_net.ipv4.ip'
```

Thoughts?

joto:
That sounds like a good solution. Not quite as easy due to the extra `jq` needed, but much more flexible.

thcyron:
I implemented JSON and Go template output in #166. Nevertheless, thanks for your PR and feedback!

PR #165, opened 2019-01-29
Allow option --user-data-from-file multiple times on "server create".

joto:
The option --user-data-from-file can now be used multiple times on a
"server create" command. If this option is specified more than once, all
files will be opened and put into a multipart MIME message which will
then be send to the API. The cloud-init can understand this MIME
message and unpack it.


thcyron:
Thank you!

PR #166, opened 2019-01-30
Add support for JSON and template output

### LINKED ISSUE #82: Structured access to fields (JSON, Go templates)

thcyron:
Closes #82 and #164.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 * Add support for executing commands via `hcloud server ssh <server> <command>`
 * Make overriding context via `HCLOUD_CONTEXT` work
+* Add support for JSON and Go template output
 
 ## v1.11.0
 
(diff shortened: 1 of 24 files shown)


joto:
Thanks. This looks pretty good. It is a bit difficult to find out which fields there are, though. Especially with the upper/lower case being important. Is there a way to output a better error message? So instead of something like:

`hcloud: template: :1:12: executing "" at <.PublicNet.IPv4.ip>: can't evaluate field ip in type hcloud.ServerPublicNetIPv4`

get something like:

`Unknown field 'ip'. Available fields in 'ServerPublicNetIPv4' are: IP, ...`

thcyron:
Adding more descriptive error messages for templates isn’t a low-hanging fruit. For now, you have to consult the [hcloud-go docs](https://godoc.org/github.com/hetznercloud/hcloud-go/hcloud) for struct layouts and available fields.

PR #168, opened 2019-02-06
Add token length validation when creating a new context.

LKaemmerling:


cli/context_create.go
```diff
@@ -49,6 +49,10 @@ func runContextCreate(cli *CLI, cmd *cobra.Command, args []string) error {
 		if token == "" {
 			continue
 		}
+		if len(token) < 64 {
```
thcyron: Shouldn’t it be exactly 64 chars long?

LKaemmerling: Yes at the moment it should, but in the future we could extend the token length to something higher. With this we don't need to change it in the CLI.

Issue #169, opened 2019-02-06
Mark active context when listing


ctron:
Issuing `hcloud context list` show the list of active contexts, it would be great if the active context would be marked in some way.

thcyron:
Implemented in https://github.com/hetznercloud/cli/pull/171.

PR #171, opened 2019-02-06
Mark active context as current

LKaemmerling:
Proposal for #169

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -15,8 +15,9 @@ type Config struct {
 }
 
 type ConfigContext struct {
-	Name  string
-	Token string
+	Name    string
+	Token   string
+	Current string
 }
 
 func (config *Config) ContextByName(name string) *ConfigContext {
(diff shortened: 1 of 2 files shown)


cli/config.go
```diff
@@ -15,8 +15,9 @@ type Config struct {
 }
 
 type ConfigContext struct {
-	Name  string
-	Token string
+	Name    string
+	Token   string
+	Current string
```
thcyron: That field doesn’t belong into the model as it’s specific to presentation.

LKaemmerling: I have changed it to a specific model for the presentation.

thcyron: 👏

thcyron:
1. Changelog entry missing
2. What about renaming “current” to ”active”?

LKaemmerling:
I have renamed it and added the Changelog entry.

PR #175, opened 2019-03-08
Add travis auto release on tag

LKaemmerling:
This allows travis to build the binaries on every tag.

So in the future we only need to change the Changelog and then create a new tag. Travis will publish the built binaries to the Github Release.

Also we will test on Go 1.11 and Go 1.12 with this PR and build with Go 1.12


thcyron:
That’s cool! Would it be possible to also automatically copy the changelog to the GitHub release? Like parsing CHANGES.md and getting the content under the `${TRAVIS_TAG}` heading?

PR #176, opened 2019-03-08
Unify Server Information on List Outputs

### LINKED ISSUE #174: Show server name instead of ID in volume list
```
$ hcloud volume list
ID        NAME                                       SIZE    SERVER    LOCATION
1234567   pvc-9e423c73-2bb3-11e9-a941-9600001a7c3e   10 GB   1234567   nbg1
```

Would be nice to have the server’s name shown instead of its ID.

LKaemmerling:
Closes #174

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -34,10 +34,9 @@ type CLI struct {
 
 func NewCLI() *CLI {
 	cli := &CLI{
-		Context:     context.Background(),
-		Config:      &Config{},
-		ConfigPath:  DefaultConfigPath,
-		serverNames: map[int]string{},
+		Context:    context.Background(),
+		Config:     &Config{},
+		ConfigPath: DefaultConfigPath,
 	}
 	if s := os.Getenv("HCLOUD_CONFIG"); s != "" {
 		cli.ConfigPath = s
@@ -205,7 +204,8 @@ func (c *CLI) WaitForActions(ctx context.Context, actions []*hcloud.Action) erro
 }
 
 func (c *CLI) GetServerName(id int) string {
-	if len(c.serverNames) == 0 {
+	if c.serverNames == nil {
+		c.serverNames = map[int]string{}
 		servers, _ := c.Client().Server.All(c.Context)
 		for _, server := range servers {
 			c.serverNames[server.ID] = server.Name
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


cli/floatingip_list.go
```diff
@@ -112,3 +66,53 @@ func runFloatingIPList(cli *CLI, cmd *cobra.Command, args []string) error {
...
+				dns = fmt.Sprintf("%d entries", len(floatingIP.DNSPtr))
+			}
+			return na(dns)
+		})).
+		AddFieldOutputFn("server", fieldOutputFn(func(obj interface{}) string {
+			floatingIP := obj.(*hcloud.FloatingIP)
+			var server string
+			if floatingIP.Server != nil && cli != nil {
+				_server, _, _ := cli.Client().Server.GetByID(cli.Context, floatingIP.Server.ID)
+				server = _server.Name
```
thcyron: ` _server` can be `nil`.

thcyron:
I wonder if it would make sense to retrieve servers in parallel instead of sequentially. Also, since API calls aren’t cheap, we should not retrieve the same server multiple times if it appears in multiple images, Floating IPs, etc.

cli/floatingip_list.go
```diff
@@ -88,7 +88,9 @@ func describeFloatingIPListTableOutput(cli *CLI) *tableOutput {
 			var server string
 			if floatingIP.Server != nil && cli != nil {
 				_server, _, _ := cli.Client().Server.GetByID(cli.Context, floatingIP.Server.ID)
-				server = _server.Name
+				if _server != nil {
+					return _server.Name
+				}
```
thcyron: Should fallback to server ID instead of n/a in case of an error (which would be wrong).

LKaemmerling:
> I wonder if it would make sense to retrieve servers in parallel instead of sequentially. Also, since API calls aren’t cheap, we should not retrieve the same server multiple times if it appears in multiple images, Floating IPs, etc.

I think I found a good solution for this. We will now fetch all servers once, when we need the first name and then use a map with the server id as key and the server name as value.

cli/cli.go
```diff
@@ -199,3 +203,16 @@ func (c *CLI) WaitForActions(ctx context.Context, actions []*hcloud.Action) erro
 
 	return nil
 }
+
+func (c *CLI) GetServerName(id int) string {
+	if len(c.serverNames) == 0 {
```
thcyron:
```suggestion
	if c.serverNames == nil {
```

Eliminates false-positives when `len(c.serverNames)` is indeed 0 because there are no servers in the project.

cli/image_list.go
```diff
@@ -132,3 +78,61 @@ func runImageList(cli *CLI, cmd *cobra.Command, args []string) error {
...
+			return fmt.Sprintf("%.0f GB", image.DiskSize)
+		})).
+		AddFieldOutputFn("created", fieldOutputFn(func(obj interface{}) string {
+			image := obj.(*hcloud.Image)
+			return humanize.Time(image.Created)
+		})).
+		AddFieldOutputFn("bound_to", fieldOutputFn(func(obj interface{}) string {
+			image := obj.(*hcloud.Image)
+			if image.BoundTo != nil && cli != nil {
+				_server, _, _ := cli.Client().Server.GetByID(cli.Context, image.BoundTo.ID)
```
thcyron: Can we use `cli.GetServerName()` here too?

PR #220, opened 2019-12-11
Add support for freebsd/arm64

dmgk:
Go 1.14 will add support for freebsd/arm64, golang.org/x/sys@33540a1f6037 is the earliest commit that brings in necessary arm64 bits.


LKaemmerling:
When Go 1.14 will be released we will add freebsd/arm64 to our builds.

dmgk:
@LKaemmerling That's fine, thanks. Just upstreaming local fixes I had to do while testing with go master branch.

PR #225, opened 2020-02-17
Add JSON output option to all hcloud list commands

### LINKED ISSUE #222: Server list as JSON
It would be nice if we could get a `hcloud server list -o json,column=name,ipv6` so we can work with the data using `jq` or python or similar… Parsing the table stable format is feasible but not really as nice to integrate in other scripts around managing a server.

LKaemmerling:
Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>

Closes #222

### DIFF
--- a/cli/network_list.go
+++ b/cli/network_list.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"github.com/hetznercloud/hcloud-go/hcloud/schema"
 	"strings"
 
 	"github.com/hetznercloud/hcloud-go/hcloud"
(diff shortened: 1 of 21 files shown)


cli/util.go
```diff
@@ -89,8 +89,10 @@ func describeFormat(object interface{}, format string) error {
 	return t.Execute(os.Stdout, object)
 }
 
-func describeJSON(object interface{}) error {
+func describeJSON(object interface{}, indent bool) error {
 	enc := json.NewEncoder(os.Stdout)
-	enc.SetIndent("", "  ")
+	if indent {
```
thcyron: What’s the reason for making this configurable?

cli/sshkey_list.go
```diff
@@ -49,6 +49,11 @@ func runSSHKeyList(cli *CLI, cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if outOpts.IsSet("json") {
+		describeJSON(sshKeys, false)
```
thcyron: Why not `return describeJSON(sshKeys, false)`, i.e. why ignore the error?

cli/network_list.go
```diff
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"github.com/hetznercloud/hcloud-go/hcloud/schema"
```
thcyron: Belongs to the second group.

PR #238, opened 2020-05-11
Add command to request a VNC console

minus7:
Add command to request a VNC console. Also includes a proxy translating WebSocket to a normal TCP socket using the `-l` flag. This allows a normal VNC viewer to connect to the console by connecting to localhost:5900.

This also updates the version of the hcloud library to make use of the console API added in https://github.com/hetznercloud/hcloud-go/pull/134

Preview:
screenshot

### DIFF
--- a/cli/server_request_console.go
+++ b/cli/server_request_console.go
@@ -16,10 +16,12 @@ func newServerRequestConsoleCommand(cli *CLI) *cobra.Command {
 		PreRunE:               cli.ensureToken,
 		RunE:                  cli.wrap(runServerRequestConsole),
 	}
+	addOutputFlag(cmd, outputOptionJSON())
 	return cmd
 }
 
 func runServerRequestConsole(cli *CLI, cmd *cobra.Command, args []string) error {
+	outOpts := outputFlagsForCommand(cmd)
 	idOrName := args[0]
 	server, _, err := cli.Client().Server.Get(cli.Context, idOrName)
 	if err != nil {
@@ -36,6 +38,16 @@ func runServerRequestConsole(cli *CLI, cmd *cobra.Command, args []string) error 
 
 	if err := cli.ActionProgress(cli.Context, result.Action); err != nil {
 		return err
+	}
+
+	if outOpts.IsSet("json") {
+		return describeJSON(struct {
+			WSSURL   string
+			Password string
+		}{
+			WSSURL:   result.WSSURL,
+			Password: result.Password,
+		})
 	}
 
 	fmt.Printf("Console for server %d:\n", server.ID)
--- a/cli/server_request_console.go
+++ b/cli/server_request_console.go
@@ -1,0 +1,45 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func newServerRequestConsoleCommand(cli *CLI) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "request-console [FLAGS] SERVER",
+		Short:                 "Request a WebSocket VNC console for a server",
+		Args:                  cobra.ExactArgs(1),
+		TraverseChildren:      true,
+		DisableFlagsInUseLine: true,
+		PreRunE:               cli.ensureToken,
+		RunE:                  cli.wrap(runServerRequestConsole),
+	}
+	return cmd
+}
+
+func runServerRequestConsole(cli *CLI, cmd *cobra.Command, args []string) error {
+	idOrName := args[0]
+	server, _, err := cli.Client().Server.Get(cli.Context, idOrName)
+	if err != nil {
+		return err
+	}
+	if server == nil {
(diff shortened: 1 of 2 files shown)


cli/server_request_console.go
```diff
@@ -0,0 +1,109 @@
...
+
+	ws, err := websocket.Dial(url, "binary", "https://console.hetzner.cloud/")
+	if err != nil {
+		return err
+	}
+	ws.PayloadType = websocket.BinaryFrame
+	fmt.Printf("Connected!\n")
+	defer ws.Close()
+
+	in := make(chan error)
```
emersion, line 51: You can make the channel buffered (`make(chan error, 1)`) to avoid leaking goroutines.

minus7: Thanks

minus7:
Bump. Rebased onto latest master. The commit for the version bump for the hcloud library is now gone. This is still good to go from my side.

LKaemmerling:
Hey @minus7,

first of all thank you for the MR. We tried it out locally (the proxy option) and on our machines (tested on MacOS and Linux) the VNC connection does not work.

I would prefer to not let the cli proxy the connection, it should just "output" the data.

minus7:
> We tried it out locally (the proxy option) and on our machines (tested on MacOS and Linux) the VNC connection does not work.

Do you have more info on why it doesn't work? I run `./hcloud server request-console -l zeus` and then `gvncviewer` with `localhost` or `localhost:0`. Note that `localhost:5900` might not work because at `gvncviewer` adds 5900 to that.

> I would prefer to not let the cli proxy the connection, it should just "output" the data.

If you don't specify `-l` you just get the data. I like having a proxy in there as a convenience tool so you don't need to find a VNC tool that supports WebSocket connection or have a separate proxy first. In fact, I'd love to have it also launch my VNC viewer and enter the VNC password for me (behind a flag of course) to minimize the steps I have to take when opening a console.

LKaemmerling:
We (@fhofherr on Linux and i  on MacOS) tested it the same way:
1. Create a server
2. Request console with -l (`hcloud server request-console <idOrName> -l`)
3. Open a random vnc client (@fhofherr used: TigerVNC Viewer 64-bit v1.10.1, i tried the native mac vnc capability and https://apps.apple.com/de/app/remote-desktop-vnc/id472995993?mt=12)
4. Enter address (tried: `localhost`, `localhost:5900`, `vnc://localhost`,`vnc://localhost:5900`)
5. Connect
6. Enter password
7. The process from 2. closes itself only with the message: `Connected!`

As you can see, adding a proxy is mostly not that easy so i would still favor to not have the proxy itself in the hcloud cli. The cli should be as small as possible.

minus7:
I guess your VNC client connects twice and the proxy exits after the first connection disconnects.

I'll remove the proxy commit then if I can't convince you to take it :)
In that case the credentials should (at least optionally) be output in a machine-readable form, so another tool can grab them. Any suggestions/preferences for that?

LKaemmerling:
It would be good if you could add the -o flag (have a look at the list endpoints) with json, so basically output it as json :)

minus7:
Proxy mode is gone now and JSON added

LKaemmerling:
Thank you! Will be available with the next release version.

sjaeckel:
I would have preferred if the proxy option would have been merged, but OK ...

FTR: connecting to the websocket URL is possible via [websocat](https://github.com/vi/websocat)

```
$ hcloud server request-console <server>
Console for server 00000:
WebSocket URL: wss://web-console.hetzner.cloud/?server_id=00000&token=abc..xyz
VNC Password: abc..xyz
$ websocat --oneshot -b tcp-l:127.0.0.1:5900 'wss://web-console.hetzner.cloud/?server_id=00000&token=abc..xyz'
```

You have to make sure to enclose the websocket URL in single-quotes `'`, otherwise your shell will most likely try to interpret some of the URL arguments and it will fail.

A one-liner of this (with some limitations like only 1 parallel instance) is

```
$ hcloud server request-console <server> | tee /dev/tty | awk -F': ' '/WebSocket/ {print $2}' | xargs -I {} websocat --oneshot -b tcp-l:127.0.0.1:5900 {}
```

Then connect with your preferred VNC client to `localhost:5900` and enter the password that was printed.

minus7:
I wrapped that up in a shell script for ease of use: `$ hcloud-vnc <server>`
Requires jq, websocat and tigervnc

```bash
#!/bin/bash

j=$(hcloud server request-console -o json "$1")

url=$(echo "$j" | jq -r '.WSSURL')
password=$(echo "$j" | jq -r '.Password')

tmpdir="$(mktemp -d ${XDG_RUNTIME_DIR:-/tmp}/hcloud-vnc.XXXXXXXXXX)"
trap 'rm -rf "$tmpdir"' EXIT

websocat --oneshot --binary unix-listen:"$tmpdir/vnc.sock" "$url" &

VNC_PASSWORD=$password vncviewer "$tmpdir/vnc.sock"
wait
```

PR #240, opened 2020-05-18
added instructions for completion with antigen in zsh.

eike-fokken:
This pull requestion updates the README, so that users of zsh and antigen can copy-paste a way to have hcloud completion working.


LKaemmerling:
Thank you!

eike-fokken:
You're welcome!

PR #262, opened 2020-08-16
Update README /w usage of enable-protection

tlrh314:
…/issues/261


LKaemmerling:
Thank you, but we think that this should not be part of the Readme. (More details in the issue).

PR #287, opened 2020-11-05
Switch bats to maintained bats-core (CI)

ktomk:
sstephenson/bats is not maintained any longer [1].

it spilled errors when running tests. replacing it with
bats-core/bats-core worked out of the box.

[1]: https://github.com/sstephenson/bats/pull/269 etc.


LKaemmerling:
Thank you!

PR #293, opened 2021-01-11
Refactor CLI commands into multiple packages

mfrister:
Having all commands in a single package makes the structure hard to navigate and understand.

This PR changes the package structure in a way that resembles the structure of the CLI commands, with one package per top-level CLI command.


LKaemmerling (approved):
Good job! I will test it in the next few days locally.

LKaemmerling (approved):

PR #370, opened 2021-12-06
rename command flag to be more explicit

lpmi-13:
the command flag `--http-redirect-http` is used to redirect traffic from port 80 to 443, so a more obvious name for the flag would be `--http-redirect-https` to be descriptive for what it does.


fhofherr:
Hi @lpmi-13,

thank you for your pull request. Unfortunately changing the flag affects other users which may be using it in scripts or similar. This would constitute a backwards incompatible change which we would like to avoid. At the same time introducing an additional flag seems a bit too much.

Therefore, while we really appreciate your contribution, we have to reject it.

PR #422, opened 2022-11-16
Update the toml library to the latest version

cedi:
github.com/pelletier/go-toml announced a new version (v2). This commit updates to v2 of said library


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/guptarohit/asciigraph v0.5.5
 	github.com/hetznercloud/hcloud-go v1.37.0
 	github.com/pelletier/go-toml v1.9.5
```
apricote, line 11: I think this line needs to be updated to `github.com/pelletier/go-toml/v2 v2.0.6`.

This happens automatically if I run `go mod tidy` locally.

apricote (changes requested):

go.mod
cedi: oh yeah. Good catch. Thanks. Fixed now

apricote (approved):

Issue #427, opened 2022-12-01
Release v1.31.0 is missing all assets


nicholasdille:
The GitHub action for the release v1.31.0 has failed before adding the release assets.

apricote:
Thanks for the report. We currently have some issues with signing the binary for macOS, which causes a timeout in the release pipeline.

I will restart the pipeline.

nicholasdille:
I hope you can get this off your back quickly #hugops

apricote:
Release `v1.31.1` was published and all assets are available: https://github.com/hetznercloud/cli/releases/tag/v1.31.1

PR #429, opened 2022-12-05
[TEST] fix(ci): find cause for #427

apricote:
We are currently trying to find the cause for #427, and this PR contains our various attempts to find the issue.

Already tested:
- macos 11
- macos 12


apricote:
It works with either `sudo gon` or by reverting the goreleaser version from `v1.13.1` to `v1.12.3`. We also realized that we did not notarize our macos execs since v1.29.3 because of this commit: https://github.com/hetznercloud/cli/commit/3c93ff63d0ea1d2fdddee730256cefedc7c4fd1b

PR #433, opened 2022-12-07
fix(ci): race-condition in signing macos binaries

apricote:
By default the keychain that is used to sign macOS binaries is only unlocked for 300s. This caused the release process to fail, as building all binaries took longer than the 300s, and subsequent signing failed.

The new timeout is 3600s (1h), which should be enough to build anything we want. At the moment building all binaries takes ~8m, so we have a buffer of 52m or 650%.

Further changes made here:

- fix(ci): show logs of gon hooks

  This makes debugging any issues encountered in these steps easier.
- fix(ci): use similar goreleaser settings to tagged release builds

  In release builds we use `--parallelism=2` and the latest goreleaser version. We should use the same settings in our dev builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,8 +36,8 @@ jobs:
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
-          version: v1.12.3
-          args: release --rm-dist --skip-publish --skip-sign
+          version: latest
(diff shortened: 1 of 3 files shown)


apricote:
Successful job: https://github.com/hetznercloud/cli/actions/runs/3639130433/jobs/6142146457

LKaemmerling (approved):

PR #455, opened 2023-02-24
chore: replace hcapi.LoadBalancerClient usages with hcapi2

samcday:


internal/cmd/loadbalancer/change_algorithm.go
```diff
@@ -1,55 +1,57 @@
...
+			ValidArgsFunction:     cmpl.SuggestArgs(cmpl.SuggestCandidatesF(client.LoadBalancer().Names)),
+			TraverseChildren:      true,
+			DisableFlagsInUseLine: true,
+		}
+
+		cmd.Flags().String("algorithm-type", "", "The new algorithm of the Load Balancer")
+		cmd.RegisterFlagCompletionFunc("algorithm-type", cmpl.SuggestCandidates(
+			string(hcloud.LoadBalancerAlgorithmTypeRoundRobin),
+			string(hcloud.LoadBalancerAlgorithmTypeRoundRobin),
+		))
```
apricote, line 30: I know this was a bug in the origin code, but we should fix this while we are at it. I can also do it in a follow up PR.

```suggestion
		cmd.RegisterFlagCompletionFunc("algorithm-type", cmpl.SuggestCandidates(
			string(hcloud.LoadBalancerAlgorithmTypeRoundRobin),
			string(hcloud.LoadBalancerAlgorithmTypeLeastConnections),
		))
```

apricote (approved):

apricote (approved):

PR #466, opened 2023-03-30
test: fix gitlab test setup

apricote:
- Remove breaking GO* env variables
- Remove dependency on internal repositories, as they are not working for us


samcday (approved):
:seal: of :+1:

PR #472, opened 2023-04-14
feat: add github action helper

mabels:
Hi,

i missed a nice way to integrate hcloud-cli into a github action. So i copied a skeleton over and make it work.

This need just a little more love e.g. a some docu and a naming review.

      - uses: mabels/hcloud-cli@add_github_action
   should be:
      - uses: hetznercloud/cli@v1.33.0

this installs the executable in your action.

Hopefully that helps.

As far as i got now the cli and the backend works like expected.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,28 @@
+name: 'Setup HCloud'
+description: 'Github Action provides the hcloud tool'
+inputs:
+  version:
+    description: 'the version of the hcloud tool'
+    required: false
+    default: 'v1.33.0'
+  url: 
+    description: 'base url to download'
+    required: false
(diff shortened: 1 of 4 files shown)


mabels:
The 11k lines are only webpack generated java script. For sure we can port this to some thing smaller but if you remove the webpacked code this is only ~100 lines.

Meno

jooola:
If such addition happens, I think the action should go in a dedicated `hetznercloud/setup-hcloud-action` repository.

jooola:
Solved with the next setup-hcloud action: https://github.com/hetznercloud/setup-hcloud

PR #478, opened 2023-05-08
ci: bump all actions to latest major version

apricote:
These actions were using Node 12, which will be shut down on 2023-05-18: https://github.blog/changelog/2023-05-04-github-actions-all-actions-will-run-on-node16-instead-of-node12/

While I was in the files, I also streamlined to workflows so they are more similar to eachother and optimized the asset uploading for preview builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,52 +11,43 @@ jobs:
     runs-on: macos-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
         with:
           go-version: 1.18
+
       - name: Print Go Version
         run: go version
+
       - name: Import GPG key
         id: import_gpg
-        uses: crazy-max/ghaction-import-gpg@v3
+        uses: crazy-max/ghaction-import-gpg@v5
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -11,52 +11,43 @@ jobs:
...
-          name: Upload Linux AMD64 Build Artifact
-          path: dist/hcloud-build_linux_amd64_v1/hcloud
-      - uses: actions/upload-artifact@v1
+
+      - uses: actions/upload-artifact@v3
         with:
-          name: Upload FreeBSD AMD64 Build Artifact
-          path: dist/hcloud-build_freebsd_amd64_v1/hcloud
+          name: Preview Binaries
+          path: dist/hcloud-*/hcloud
```
apricote, line 53: This simplifies the workflow file and now all built binaries are available. The downside is that the user needs to download all binaries as one zip, instead of one-zip-per-binary. Not sure if this tradeoff is worth it.

samcday: If this is just for pre-releases then it seems fine to me.

samcday (approved):
Renovate has been picking these version bumps up for us elsewhere. Should we be considering running it here?

apricote:
> Renovate has been picking these version bumps up for us elsewhere. Should we be considering running it here?

Yes! Though with the major versions we still need to adjust some things here and there anyway. See _INTEGRATIONS-13_ on the internal issue tracker.

PR #482, opened 2023-05-31
fix(server): wait for delete before returning

apricote:
The `hcloud server delete` command returned before the action was actually awaited. This caused a race condition when creating new servers with the same name.


DrBu7cher (approved):
:+1:

PR #504, opened 2023-06-21
feat(network): add support for exposing routes to vswitch connection

apricote:
Feature was released today. This adds:

- Showing the field in `network describe` (also for `-o json`)
- Setting the field through flag on `network create`
- Updating the setting through the command `network expose-routes-to-vswitch`

### DIFF
--- a/internal/cmd/network/create.go
+++ b/internal/cmd/network/create.go
@@ -26,6 +26,8 @@ func newCreateCommand(cli *state.State) *cobra.Command {
 	cmd.Flags().IPNet("ip-range", net.IPNet{}, "Network IP range (required)")
 	cmd.MarkFlagRequired("ip-range")
 
+	cmd.Flags().Bool("expose-routes-to-vswitch", false, "Expose routes from this network to the vSwitch connection. It only takes effect if a vSwitch connection is active.")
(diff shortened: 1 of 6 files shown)


jooola (approved):
Aren't we forgetting the update command ?

apricote:
> Aren't we forgetting the update command ?

Instead of adding the field to `hcloud network update` I added the new command `hcloud network expose-routes-to-vswitch`. This way its easier to add some more documentation to it, as its a potentially dangerous option.

Previously we mostly added new subcommands for actions. The only subcommands the use the general "update" api endpoints are for managing labels. Do you think its easier for the user to understand if its part of the `hcloud network update` command?

jooola:
> Previously we mostly added new subcommands for actions. The only subcommands the use the general "update" api endpoints are for managing labels. Do you think its easier for the user to understand if its part of the hcloud network update command?

I think it's about being consistent, I would assume that if I have an update command, it will be consistent with the rest of the integrations. Either we document that the update command does not include all the fields, or we add multiple ways of updating such values.

apricote:
I added a note to the `hcloud network update` command to send users the right way.

PR #530, opened 2023-08-15
feat: allow formatting a volume without automounting it

### LINKED ISSUE #529: Formatting a volume without automounting is not possible
When creating a volume, the following options are available:

```
...
      --automount              Automount volume after attach (server must be provided)
      --format string          Format volume after creation (automount must be enabled)
...
```

It states that the `format` option must not be used without `--automount`. However, the REST API does not have that constraint:

<details><summary>See curl output…</summary>

```bash
curl \
        -X POST \
        -H "Authorization: Bearer $API_TOKEN" \
        -H "Content-Type: application/json" \
        -d '{"automount":false,"format":"xfs","location":"fsn1","name":"tmp","size":10}' \
        'https://api.hetzner.cloud/v1/volumes'
{
  "volume": {
    "id": 35727348,
    "created": "2023-08-15T18:38:14+00:00",
    "name": "tmp",
    "server": null,
    "location": {
      "id": 1,
      "name": "fsn1",
      "description": "Falkenstein DC Park 1",
      "country": "DE",
      "city": "Falkenstein",
      "latitude": 50.47612,
      "longitude": 12.370071,
      "network_zone": "eu-central"
    },
    "size": 10,
    "linux_device": "/dev/disk/by-id/scsi-0HC_Volume_35727348",
    "protection": {
      "delete": false
    },
    "labels": {},
    "status": "creating",
    "format": "xfs"
  },
  "action": {
    "id": 1127437523,
    "command": "create_volume",
    "status": "running",
    "progress": 0,
    "started": "2023-08-15T18:38:14+00:00",
    "finished": null,
    "resources": [
      {
        "id": 35727348,
        "type": "volume"
      }
    ],
    "error": null
  },
  "next_actions": []
}
```

</details>

Using `hcloud volume attach --server=tmp --automount tmp`, that volume can be attached to the server and is formatted already:

```
/dev/sdb       xfs     10G  104M  9.9G   2% /mnt/HC_Volume_35727348
```

When running the equivalent command using the CLI, the volume is created, but the `format` option is silently ignored and an unformatted volume is created.

---

Additionally, it might be helpful to document the supported format strings both in the `--help` output and via shell completion for the `--format` option.

marvinruder:
* Fixes #529

### To Do
- [x] document supported volume formats (this can be done either hard-coded or fetched from somewhere, but that may require changing the REST API service, so I left that open for now)

### DIFF
--- a/internal/cmd/volume/create.go
+++ b/internal/cmd/volume/create.go
@@ -35,7 +35,7 @@ var CreateCommand = base.Cmd{
 		cmd.MarkFlagRequired("size")
 
 		cmd.Flags().Bool("automount", false, "Automount volume after attach (server must be provided)")
-		cmd.Flags().String("format", "", "Format volume after creation (automount must be enabled)")
+		cmd.Flags().String("format", "", "Format volume after creation")
 
 		cmd.Flags().StringToString("label", nil, "User-defined labels ('key=value') (can be specified multiple times)")
 
@@ -76,9 +76,9 @@ var CreateCommand = base.Cmd{
 		}
 		if automount {
 			opts.Automount = &automount
-			if format != "" {
-				opts.Format = &format
-			}
+		}
+		if format != "" {
+			opts.Format = &format
 		}
 
 		result, _, err := client.Volume().Create(ctx, opts)
--- a/internal/cmd/volume/create.go
+++ b/internal/cmd/volume/create.go
@@ -35,7 +35,9 @@ var CreateCommand = base.Cmd{
 		cmd.MarkFlagRequired("size")
 
 		cmd.Flags().Bool("automount", false, "Automount volume after attach (server must be provided)")
-		cmd.Flags().String("format", "", "Format volume after creation")
+
+		cmd.Flags().String("format", "", "Format volume after creation (ext4 or xfs)")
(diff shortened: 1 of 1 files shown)


internal/cmd/volume/create.go
```diff
@@ -35,7 +35,7 @@ var CreateCommand = base.Cmd{
 		cmd.MarkFlagRequired("size")
 
 		cmd.Flags().Bool("automount", false, "Automount volume after attach (server must be provided)")
-		cmd.Flags().String("format", "", "Format volume after creation (automount must be enabled)")
+		cmd.Flags().String("format", "", "Format volume after creation")
```
apricote, line 38: Regarding your TODO, I think we can just hardcode them, we did something similar for the `network add-subnet --type` flag. If we ever add another format, the user can still pass it in because we dont validate against the autocompletion.

```suggestion

		cmd.Flags().String("format", "", "Format volume after creation (ext4 or xfs)")
		cmd.RegisterFlagCompletionFunc("format", cmpl.SuggestCandidates("ext4", "xfs"))
```

apricote (commented):
Nice change! You are right, it makes no sense to restrict formatting to automounted volumes.

apricote (approved):

PR #543, opened 2023-09-15
feat: build with Go 1.21

apricote:
Also bump golangci-lint to v1.54.2 and fix warnings:

- Linters `structcheck`, `deadcode` & `varcheck` are deprecated by their upstream


.github/workflows/ci.yml
```diff
@@ -1,5 +1,13 @@
 name: Continuous Integration
-on: [push, pull_request]
+
+on:
+  push:
+    tags:
+      - v*
```
jooola, line 6: I am not sure if this is really necessary.

jooola (approved):

.github/workflows/ci.yml
apricote: Probably not, we do however do this in many of our repos, and the current workflow also runs for tags.

Issue #489, opened 2023-06-05
fix: hcloud server shutdown should wait until the server is actually shut down


apricote:
Running `hcloud server shutdown` only waits until the ACPI shutdown signal is sent to the server, not until the server is actually shut down. This can cause issues if the user is directly doing another action right after (like unassigning Primary IPs).

This is expected based on the action returned by the API.

We should poll the server to wait until it is actually marked as shutdown in the API to prevent race conditions.

pauhull:
Implemented in #569

PR #569, opened 2023-10-19
feat(server): add --wait flag to "shutdown" command

phm07:
As described in #489

### DIFF
--- a/internal/cmd/server/shutdown.go
+++ b/internal/cmd/server/shutdown.go
@@ -2,7 +2,10 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+	"time"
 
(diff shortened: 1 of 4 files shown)


internal/cmd/server/shutdown.go
```diff
@@ -41,7 +60,46 @@ var ShutdownCommand = base.Cmd{
...
+
+			go func() {
+				defer close(errCh)
+
+				ticker := time.NewTicker(interval)
+				defer ticker.Stop()
+
+				for server.Status != hcloud.ServerStatusOff {
+					if now := <-ticker.C; now.Sub(start) >= timeout {
+						fmt.Println()
```
apricote, line 82: What is the purpose of this new line?

apricote (commented):
Very Nice :rocket: Only got one question/nit

apricote (approved):

PR #577, opened 2023-10-23
docs(load-balancer): change description for protocol flag

n9v9:
This fixes the description for the `protocol` flag of the `load-balancer update-service` command. Before this change, the description was identical with the description of the `health-check-protocol` flag.


apricote (approved):

apricote:
Thank you very much @n9v9 :)

Issue #545, opened 2023-09-29
feat: Command to list all resources


apricote:
Introduce a new command `hcloud all list` that lists all resources in the project. This command should support label selectors. The default output would be similar to `kubectl get all`, one table per resource type:

```
$ kubectl get all -n traefik
NAME                           READY   STATUS    RESTARTS       AGE
pod/traefik-56fc75b747-fxd4k   1/1     Running   1 (6d3h ago)   6d3h

NAME              TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)                      AGE
service/traefik   LoadBalancer   10.245.208.234   10.0.0.7      80:32466/TCP,443:30059/TCP   198d

NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/traefik   1/1     1            1           198d

NAME                                 DESIRED   CURRENT   READY   AGE
replicaset.apps/traefik-56fc75b747   1         1         1       18d
replicaset.apps/traefik-5dbc8db98    0         0         0       198d
replicaset.apps/traefik-6f9987db55   0         0         0       116d
```

Open questions:

- How do we implement the `-o=columns=` and `-o=json` flags from the usual `list` command?

stroebitzer:
I would need this also, additonally I would like to filter the resources by the API token name with which the resources have been created.

pauhull:
Implemented in #580

stroebitzer:
As far as I got it the API-token name is not included in the provided info. Do you intend adding this info?

pauhull:
Unfortunately, we can only implement features supported by the Hetzner Cloud Public API (https://docs.hetzner.cloud/).
API tokens are not included in the Public API. It also does not expose who created which resource.

However, you can find this information on the activities page in the Cloud Console: https://console.hetzner.cloud/activities

I have added a vote for this feature to our internal feedback tracker.

stroebitzer:
thx for your response. 

to give you some context about the reason why we are interested in those features. Maybe there is also a better way for supporting our situation. We are doing trainings on your VMs, everything has to be done in an automated way, so no UI involved on setup and teardown.

The easiest for us would be creation and deletion of the training projects (and all its resources), but this is not supported AFAIK

So keeping the project and doing deletion of resources identified by the API token name would my second choice TBH  :)

apricote:
[...]

PR #580, opened 2023-10-24
feat: add "all list" command to list all available resources

pauhull:
As described in #545

### DIFF
--- a/internal/cmd/all/list.go
+++ b/internal/cmd/all/list.go
@@ -1,0 +1,186 @@
+package all
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/hetznercloud/cli/internal/cmd/base"
+	"github.com/hetznercloud/cli/internal/cmd/certificate"
+	"github.com/hetznercloud/cli/internal/cmd/firewall"
+	"github.com/hetznercloud/cli/internal/cmd/floatingip"
+	"github.com/hetznercloud/cli/internal/cmd/image"
+	"github.com/hetznercloud/cli/internal/cmd/iso"
+	"github.com/hetznercloud/cli/internal/cmd/loadbalancer"
+	"github.com/hetznercloud/cli/internal/cmd/network"
+	"github.com/hetznercloud/cli/internal/cmd/output"
+	"github.com/hetznercloud/cli/internal/cmd/placementgroup"
+	"github.com/hetznercloud/cli/internal/cmd/primaryip"
+	"github.com/hetznercloud/cli/internal/cmd/server"
+	"github.com/hetznercloud/cli/internal/cmd/sshkey"
+	"github.com/hetznercloud/cli/internal/cmd/volume"
+	"github.com/hetznercloud/cli/internal/hcapi2"
+	"github.com/hetznercloud/cli/internal/state"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+)
+
+var allCmds = []base.ListCmd{
+	server.ListCmd,
+	image.ListCmd,
+	placementgroup.ListCmd,
+	primaryip.ListCmd,
+	iso.ListCmd,
+	volume.ListCmd,
+	loadbalancer.ListCmd,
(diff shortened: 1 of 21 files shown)


internal/cmd/all/list.go
```diff
@@ -0,0 +1,157 @@
...
+var listCmd = base.Cmd{
+	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
+		cmd := &cobra.Command{
+			Use:   "list FLAGS",
+			Short: "List all resources",
+		}
+
+		cmd.Flags().StringP("selector", "l", "", "Selector to filter by labels")
+
+		cmd.Flags().Bool("paid", false, "Only list paid resources")
```
apricote, line 40: Maybe this is easier to understand?

```suggestion
		cmd.Flags().Bool("paid", false, "Only list resources that cost money")
```

jooola, line 40: If I may shim in here: what does `paid` refer to ? Is that to exclude resources that were not created by the user ? e.g. official images/isos ?

internal/cmd/all/list.go
```diff
@@ -0,0 +1,157 @@
...
+
+		for i, lc := range cmds {
+			i, lc := i, lc
+			ch := make(chan response)
+			responseChs[i] = ch
+
+			go func() {
+				defer close(ch)
+
+				labelSelector, _ := cmd.Flags().GetString("selector")
```
apricote, line 94: Could be moved to line 49 to avoid reparsing the argument for every resource

internal/cmd/all/list.go
```diff
@@ -0,0 +1,157 @@
...
+				flagSet := pflag.NewFlagSet(lc.JSONKeyGetByName, pflag.ExitOnError)
+
+				switch lc.JSONKeyGetByName {
+				case image.ListCmd.JSONKeyGetByName:
+					flagSet.StringSlice("type", []string{"backup", "snapshot"}, "")
+				case iso.ListCmd.JSONKeyGetByName:
+					flagSet.StringSlice("type", []string{"private"}, "")
+				}
+
+				_ = flagSet.Parse([]string{})
```
apricote, line 108: Can you add a comment why this is necessary? To the unknowing reader this looks like a no-op.

internal/cmd/all/list.go
```diff
@@ -0,0 +1,157 @@
...
+				result, err := lc.Fetch(ctx, client, flagSet, listOpts, []string{})
+				ch <- response{result, err}
+			}()
+		}
+
+		resources := make([][]any, len(cmds))
```
apricote, line 114:
```suggestion
    // Wait for all requests to finish and collect results
		resources := make([][]any, len(cmds))
```

apricote (changes requested):

internal/cmd/all/list.go
jooola: I think introducing the notion of paid with this flag might lead to confusions.

I would rather use another flag name (e.g --owned) or invert it with --include-read-only-resource (yeah this is too long, but maybe you folks have an idea?). Using a --all flag might be a bit repetitive `all list --all`.

apricote: What would be the usecase to list all static resources? I think the user would rather like to see their resources.

---

I think we are conflating two discussions:

- _Paid_ (servers, IPs, LBs, Volumes) vs. _free_ (Network, SSH Key, Placement Group)
- _Project Resouces_ (Server, LB, PG, Snapshots+Backups Images) vs _Static/Public_ (System+App Images, Public ISOs, Location, Network Zone, Server Types, Load Balancer Types)

apricote: We have a similar pattern in `console.hetzner.cloud` on the Project Card View. Here it shows all paid resources as icons for every project and unpaid resources are grouped under "Other":

screenshot-paid-unpaid

jooola: Oh alright, so list all is only for project resources ? In that case, the paid flag name make more sense. But this isn't This wasn't clear from the current description of the command, I would have expected to list all resource (+ static/public)

pauhull: The long description now explains the command in more detail and explicitly states that static and public resources are not included.

apricote (approved):

PR #594, opened 2023-10-31
fix(primary-ip): list does now allow to output IP column

### LINKED ISSUE #591: `primary-ip list` can't select ip column
I may be missing something, but the primary-ip list subcommand does not appear to have an ip column to output when using -o columns=...
```
$ hcloud primary-ip list --help
Displays a list of Primary IPs.                      
                                                           
Output can be controlled with the -o flag. Use -o noheader to suppress the
table header. Displayed columns and their order can be set with
-o columns=age,assignee (see available columns below).                                                                                                                                                                                        
                                                                                                                                                                                                                                              
Columns:                                                                                                                                                                                                                                      
 - age                                                                                                                                                                                                                                        
 - assignee                                                                                                                                                                                                                                   
 - assignee_id                                                                                                                                                                                                                                
 - assignee_type                                                                                                                                                                                                                              
 - auto_delete                                                                                                                                                                                                                                
 - blocked                                                                                                                                                                                                                                    
 - created                                                                                                                                                                                                                                    
 - dns                                                                                                                                                                                                                                        
 - id                                                                                                                                                                                                                                         
 - labels                                                                                                              
 - name
 - protection                                                                                                                                                                                                                                 
 - type

Usage:
  hcloud primary-ip list [FlAGS]

```

pauhull:
Since IP is a ``*hcloud.PrimaryIP``, ``AddAllowedFields()`` only adds columns for primitive types and a field function for IP was missing, you were not able to output the IP column using output flags.

Closes #591

### DIFF
--- a/internal/cmd/primaryip/list.go
+++ b/internal/cmd/primaryip/list.go
@@ -38,6 +38,14 @@ var ListCmd = base.ListCmd{
 	OutputTable: func(client hcapi2.Client) *output.Table {
 		return output.NewTable().
 			AddAllowedFields(hcloud.PrimaryIP{}).
+			AddFieldFn("ip", output.FieldFn(func(obj interface{}) string {
+				primaryIP := obj.(*hcloud.PrimaryIP)
+				// Format IPv6 correctly
+				if primaryIP.Network != nil {
+					return primaryIP.Network.String()
+				}
+				return primaryIP.IP.String()
+			})).
 			AddFieldFn("dns", output.FieldFn(func(obj interface{}) string {
 				primaryIP := obj.(*hcloud.PrimaryIP)
 				var dns string
--- a/internal/cmd/floatingip/list.go
+++ b/internal/cmd/floatingip/list.go
@@ -65,6 +65,7 @@ var ListCmd = base.ListCmd{
 			})).
 			AddFieldFn("ip", output.FieldFn(func(obj interface{}) string {
 				floatingIP := obj.(*hcloud.FloatingIP)
+				// Format IPv6 correctly
 				if floatingIP.Network != nil {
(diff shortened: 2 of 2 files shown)


internal/cmd/primaryip/list.go
```diff
@@ -38,6 +38,13 @@ var ListCmd = base.ListCmd{
 	OutputTable: func(client hcapi2.Client) *output.Table {
 		return output.NewTable().
 			AddAllowedFields(hcloud.PrimaryIP{}).
+			AddFieldFn("ip", output.FieldFn(func(obj interface{}) string {
+				primaryIP := obj.(*hcloud.PrimaryIP)
+				if primaryIP.Network != nil {
+					return primaryIP.Network.String()
+				}
```
jooola, line 45:
```suggestion
				// .Network is set when .Type == "ipv6"
				if primaryIP.Network != nil {
					return primaryIP.Network.String()
				}
```

This seems to hide some logic, and doesn't make the code obvious. Feel free to add my comment (maybe also in the floating IP implementation?) if you think it makes it more understandable.

Maybe having a `.Type == "ipv6"` condition instead of `.Network != nil` would be more explicit, but lets not diverges from the original floating IP implementation.

jooola (approved):

internal/cmd/primaryip/list.go
pauhull: Yes, I think a comment would be useful. I'll add that. I don't think changing ``.Network != nil`` to ``.Type == "ipv6"`` would be good idea though because that would make a nil dereference possible.

jooola: Yes agreed.

jooola (approved):

PR #607, opened 2023-11-09
test: capture stderr for tests

pauhull:
In the future this will be useful for testing "-o=json" on resource creation, because the actual JSON and status messages will be printed to stdout/stderr respectively.

### DIFF
--- a/internal/cmd/all/list_test.go
+++ b/internal/cmd/all/list_test.go
@@ -173,7 +173,7 @@ func TestListAll(t *testing.T) {
 			},
 		}, nil)
 
-	out, err := fx.Run(cmd, []string{})
+	out, _, err := fx.Run(cmd, []string{})
 
 	expOut := `SERVERS
 ---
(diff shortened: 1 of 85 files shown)


jooola (approved):
Have you considered renaming the `out` var to `stdout` in this PR as well ?

You will probably end up with `out, stderr, err := fx.Run(cmd, []string{})` so you might as well rename everything now to maintain a consistency:
```go
stdout, _, err := fx.Run(cmd, []string{})
```

pauhull:
> Have you considered renaming the `out` var to `stdout` in this PR as well ?
> 
> You will probably end up with `out, stderr, err := fx.Run(cmd, []string{})` so you might as well rename everything now to maintain consistency:
> 
> ```go
> stdout, _, err := fx.Run(cmd, []string{})
> stdout, stderr, err := fx.Run(cmd, []string{})
> ```

In the vast majority of cases there won't be any output to stderr anyway so the second string can be safely ignored. In that case `out` is still a concise variable name in my opinion.
If you use the second return value though, I agree that the naming you suggested is sensible. I don't think renaming every `out` now would be needed though.

PR #609, opened 2023-11-09
feat: allow JSON output on resource creation

### LINKED ISSUE #470: Allow getting JSON on server creation
I am bash scripting creating a Hetzner Server and getting the IP. Right now I have to do 2 steps to create and get the ip as the create command does not have an JSON output

```
hcloud server create --image docker-ce --location fsn1 --type cpx31 --name=docker-local --ssh-key 7588395
SERVER_IP=$(hcloud server describe docker-local -o json | jq .public_net.ipv4.ip -r)
```

I would rather have in create already the server ip as JSON to save the second command :)

pauhull:
This PR adds the possibility to use the `-o=json` flag on resource creation and also adds the corresponding tests.

Closes #470


internal/testutil/testing.go
```diff
@@ -2,9 +2,16 @@ package testutil
...
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
+	"net/http/httptest"
 	"os"
+	"testing"
+
+	"gotest.tools/assert"
```
apricote, line 12: Please use `github.com/stretchr/testify` instead for fancy assertations. It is already installed and used in other tests.

internal/cmd/base/create.go
```diff
@@ -0,0 +1,79 @@
...
+
+		if isJson {
+			bytes, _ := io.ReadAll(response.Body)
+
+			var data map[string]any
+			if err := json.Unmarshal(bytes, &data); err != nil {
+				return err
+			}
+
+			delete(data, "action")
```
apricote, line 69: Some create commands also return `actions` or `nextActions`, do these also need to be deleted?

apricote (commented):
Way cleaner than my solution :rocket:

internal/cmd/base/create.go
```diff
@@ -0,0 +1,79 @@
...
+	PrintResource    func(context.Context, hcapi2.Client, *cobra.Command, any)
+}
+
+// CobraCommand creates a command that can be registered with cobra.
+func (cc *CreateCmd) CobraCommand(
+	ctx context.Context, client hcapi2.Client, tokenEnsurer state.TokenEnsurer, actionWaiter state.ActionWaiter,
+) *cobra.Command {
+	cmd := cc.BaseCobraCommand(client)
+
+	output.AddFlag(cmd, output.OptionJSON())
```
apricote, line 31: Could you also add it to the docs: https://github.com/hetznercloud/cli?tab=readme-ov-file#output-configuration

internal/testutil/testing.go
pauhull: As it turns out, `github.com/stretchr/testify` already has a method to compare JSON (oops). Implemented in 1fa7556

internal/cmd/base/create.go
pauhull: Of course they do. Implemented in 1fa7556 as well.

internal/cmd/base/create.go
pauhull: Done in 8e06561252d1ac6306abb778fd542db32b243f9b

apricote (approved):
:rocket:

PR #610, opened 2023-11-09
feat: pretty-print JSON output

pauhull:
This PR makes the JSON output (for example when using `... describe -o=json`) properly formatted, with 2 spaces as indent.


apricote (approved):
~~`{"nice":true}`~~

```json
{
  "nice": true
}
```

PR #611, opened 2023-11-14
ci: add Codecov to check test coverage

pauhull:


apricote (changes requested):
Could we add this to our existing test runs? https://github.com/hetznercloud/cli/blob/1143e072f9e186bada933fc049ea5be1d40e5a73/.github/workflows/ci.yml#L43

apricote (approved):

jooola:
If I remember correctly, codecov will fail the pipelines if the coverage status checks fail, e.g. coverage is less than 80%.

Maybe we want to wait a bit before we enable such check to make such we actually have the required coverage.

See:
https://docs.codecov.com/docs/commit-status#project-status
https://docs.codecov.com/docs/commit-status#patch-status

Also, we might need to enable the carryforward flags, but I am not sure if we have split tests in the CLI (unit/e2e) ?

https://docs.codecov.com/docs/carryforward-flags


Maybe we can start with `codecov.yml`:
```yml
---
coverage:
  status:
    project: off
    patch: off
```

jooola (approved):

PR #614, opened 2023-11-16
test: use client interfaces provided by hcloud-go for mocking

pauhull:
Once https://github.com/hetznercloud/hcloud-go/pull/342 is merged, `hcloud-go` provides its own interfaces for clients. We can thus remove the ones we generated ourselves and use the provided ones instead. Also, the `interfacer` tool can be removed.


apricote (commented):
Looking good, pending the merge in `hcloud-go` :)

apricote (approved):

PR #622, opened 2023-12-07
fix: make JSON/YAML outputs consistent

### LINKED ISSUE #461: Inconsistent resource objects when output format is json
When setting the output format to json, the resource object is rendered differently between the corresponding `list` and `describe` command. 

The following json output is just an example, this behaviour is not limited to the iso resource.
```
# list all isos
[{"id":1,"name":"dvd_a.iso","description":"DVD A","type":"","deprecated":"0001-01-01T00:00:00Z"},{"id":2,"name":"","description":"cd_b.iso","type":"","deprecated":"0001-01-01T00:00:00Z"}]

# describe iso 1
{"deprecated":null,"description":"DVD A","id":1,"name":"dvd_a.iso","type":"public"}

# describe iso 2
{"deprecated":null,"description":"cd_b.iso","id":2,"name":null,"type":"private"}
```

After digging a bit into this, I noticed that a [ListCmd](https://github.com/hetznercloud/cli/blob/main/internal/cmd/iso/list.go#L38-L45) has a `JSONSchema` attribute which converts the resource into a [hcloud-go api schema object](https://github.com/hetznercloud/hcloud-go/tree/main/hcloud/schema) before decoding it into a JSON string. In contrast the `DescribeCmd` just decodes the resource object directly.

pauhull:
This PR utilizes the new `hcloud -> schema` conversion methods introduced in https://github.com/hetznercloud/hcloud-go/pull/343 to reduce boilerplate code and enforce consistent schema conversions (in combination with https://github.com/hetznercloud/hcloud-go/pull/349 & https://github.com/hetznercloud/hcloud-go/pull/348)

Fixes #461


apricote:
Looks very nice already! I noticed that all `JSONOutput` methods are now basically the same, besides the method they are calling. Could we extract this into `ListCmd` to reduce the individual code even further?

pauhull:
I don't think this is possible. You would need generic structs and generic type casts, none of which is possible in Go as of now.

apricote (approved):
Could you rename this PR to change schema to `json/yaml`? Not all users know the implementation details

PR #628, opened 2023-12-12
feat: allow YAML output

pauhull:
This PR adds support for the `-o=yaml` flag wherever `-o=json` is possible. It uses https://github.com/goccy/go-yaml, which uses the already existing JSON tags on schema structs, so they don't need to be modified.


apricote (changes requested):
Could you also add to the the describe base command? It currently allows JSON but not YAML output.

apricote (approved):

PR #650, opened 2023-12-19
fix: integer fields are displayed as floats when using describe with -o=yaml

### LINKED ISSUE #638: fix: integer fields are displayed as floats when using `describe` with `o=yaml`
When outputting resources in YAML format using `describe`, some integer fields (like IDs) are interpreted as floating point numbers and are thus unusable. Example output of `$ hcloud image describe fedora-39 -o=yaml`:
```yaml
architecture: x86
bound_to: null
created: 2023-11-09T06:19:32+00:00
created_from: null
deleted: null
deprecated: null
description: Fedora 39
disk_size: 5.0
id: 1.35075974e+08 # interpreted as floating point
image_size: null
labels: {}
name: fedora-39
os_flavor: fedora
os_version: "39"
protection:
  delete: false
rapid_deploy: true
status: available
type: system
```

This is because when running `describe`, the API response is decoded into a `map[string]any`. The Go `json` library defaults to `float64` when converting JSON number literals into an `any`, so in the encoding back to YAML these integer fields are interpreted as floats.

pauhull:
Closes #638

### DIFF
--- a/internal/cmd/base/describe.go
+++ b/internal/cmd/base/describe.go
@@ -2,9 +2,7 @@ package base
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"io"
 	"reflect"
 	"strings"
 
@@ -15,7 +13,6 @@ import (
 	"github.com/hetznercloud/cli/internal/cmd/util"
 	"github.com/hetznercloud/cli/internal/hcapi2"
 	"github.com/hetznercloud/cli/internal/state"
-	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 )
 
 // DescribeCmd allows defining commands for describing a resource.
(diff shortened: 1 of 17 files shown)


internal/cmd/base/describe.go
```diff
@@ -27,7 +24,7 @@ type DescribeCmd struct {
 	JSONKeyGetByName string // e.g. "servers"
 	NameSuggestions  func(client hcapi2.Client) func() []string
 	AdditionalFlags  func(*cobra.Command)
-	Fetch            func(ctx context.Context, client hcapi2.Client, cmd *cobra.Command, idOrName string) (interface{}, *hcloud.Response, error)
+	Fetch            func(ctx context.Context, client hcapi2.Client, cmd *cobra.Command, idOrName string) (interface{}, interface{}, error)
```
jooola, line 27: Could we add a comment to explain what the 2 interfaces are?

jooola (changes requested):

internal/cmd/base/describe.go
jooola: And maybe we can create a `DescribeCmdFetchFunc` type to document the function and reduce the duplication for all the describe commands ?

```go
// DescribeCmdFetchFunc TODO
type DescribeCmdFetchFunc func(ctx context.Context, client hcapi2.Client, cmd *cobra.Command, idOrName string) (interface{}, interface{}, error)
```

jooola (approved):

internal/cmd/base/describe.go
pauhull: Added in 827819c9ce89f2db6dae0693f5c0d2adb83eceaa

PR #651, opened 2023-12-19
fix: integer fields are displayed as floats when using create with o=yaml

### LINKED ISSUE #638: fix: integer fields are displayed as floats when using `describe` with `o=yaml`

pauhull:
Continuation #650, closes #638

When outputting a resource after creation, it is first converted into a schema struct. This fixes bugs like #638 and also makes all schema outputs (describe, list, create) consistent.


jooola:
The json output for the create commands has already been released with 1.40.0, I think its a bit late to introduce a breaking change. Could we use a workaround to fix this ? Maybe using inline structs ?

pauhull:
Yeah, I'll add that

jooola (approved):

jooola:
:rocket:

PR #653, opened 2023-12-21
ci: prefer `go-version-file` option for `actions/setup-go`

wheinze:
In order to automatically align the Go version with the one used in development, I would suggest to prefer the `go-version-file` option of the setup action, so the Go version  does not need to be managed manually. 

PS: It might be meaningful to think of introducing a test-matrix for `.github/workflows/ci.yml`, so the officially supported Go versions are checked. If this is something you consider to do, I'm willing to file a subsequent PR for this ;)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: "go.mod"
+          go-version-file: go.mod
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v5
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version-file: "go.mod"
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version-file: "go.mod"
```
jooola, line 18:
```suggestion
          go-version-file: go.mod
```

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version-file: "go.mod"
```
jooola, line 18:
```suggestion
          go-version-file: go.mod
```

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
...
```
jooola, line 40:
```suggestion
          go-version-file: go.mod
```

.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version-file: "go.mod"
```
jooola, line 18:
```suggestion
          go-version-file: go.mod
```

jooola (commented):
Thanks for the improvement, this change does make sens in the context of this project.

The CLI does have the same go compatibility constraints as libraries, so I prefer not to introduce a test matrix for different version of go.

Thanks

jooola (approved):

wheinze:
> The CLI doesn't have the same go compatibility constraints as libraries, so I prefer not to introduce a test matrix for different version of go for now.

Alright :)

PR #654, opened 2023-12-22
ci: assign version using release-please

### LINKED ISSUE #636: ci: update version number string with release please
At the moment, the version number string in `internal/version/version.go` is updated via compiler flags during the CI build. This should be changed so that Release Please updates this string for each version. This would make `hcloud version` return the correct version number instead of `Unknown` when it's installed with `go install`.

This is also already done in other projects, see https://github.com/hetznercloud/hcloud-go/blob/main/hcloud/hcloud.go#L5

pauhull:
This PR changes version number output to the following:

### Release
```
$ hcloud version
hcloud 1.23.4
```

### goreleaser --snapshot
```
$ hcloud version
hcloud 1.23.4-dev+<short commit id>
```

### go build/go install
```
$ hcloud version
hcloud 1.23.4-dev
```

Closes #636

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -16,3 +16,6 @@ jobs:
           token: ${{ secrets.HCLOUD_BOT_TOKEN }}
           config-file: .github/release-please-config.json
           manifest-file: .github/release-please-manifest.json
+
+          extra-files: |
+            version/version.go
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -16,6 +16,3 @@ jobs:
           token: ${{ secrets.HCLOUD_BOT_TOKEN }}
           config-file: .github/release-please-config.json
           manifest-file: .github/release-please-manifest.json
-
-          extra-files: |
-            version/version.go
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,7 +7,9 @@ builds:
     main: ./cmd/hcloud/main.go
     binary: hcloud
     ldflags:
-      - -w -X github.com/hetznercloud/cli/internal/version.Version={{.Version}}
+      - -w
+      - -X {{ .ModulePath }}/internal/version.Version={{ .Version }}
+      - -X {{ .ModulePath }}/internal/version.VersionPrerelease={{- if .IsSnapshot -}}dev+{{ .ShortCommit }}{{- end -}}
     env:
       - CGO_ENABLED=0
     goos:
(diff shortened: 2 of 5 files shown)


.goreleaser.yml
```diff
@@ -27,7 +29,9 @@ builds:
     main: ./cmd/hcloud/main.go
     binary: hcloud
     ldflags:
-      - -w -X github.com/hetznercloud/cli/internal/version.Version={{.Version}}
+      - -w
+      - -X {{ .ModulePath }}/internal/version.Version={{ .Version }}
```
pauhull, line 33: Not quite sure if this is needed. I think we can remove it since release-please sets the version variable anyway.

https://github.com/hetznercloud/packer-plugin-hcloud/blob/8a95e8f92abb326cddada38da8472bb7425e9b0e/.goreleaser.yml#L17

.github/workflows/release-please.yml
```diff
@@ -16,3 +16,6 @@ jobs:
           token: ${{ secrets.HCLOUD_BOT_TOKEN }}
           config-file: .github/release-please-config.json
           manifest-file: .github/release-please-manifest.json
+
+          extra-files: |
+            version/version.go
```
jooola, line 21: This must go in the `.github/release-please-config.json` file now that we use the release-please action v4.

See https://github.com/hetznercloud/hcloud-python/blob/984022fd3888ef856be83de82554d55a8af18dba/.github/release-please-config.json#L10

pauhull: Oh right. The path is wrong anyway, it's missing `internal/`

.goreleaser.yml
jooola: True, but let's keep it so its more explicit.

pauhull: Could there be a conflict if release-please builds a new version but the tag is not yet there? Or does release-please always add the tag first?

internal/version/version.go
```diff
@@ -1,4 +1,23 @@
...
+
+	// FullVersion is the full version string, including the prerelease tag.
+	// This is dynamically generated based on the Version and Tag variables.
+	FullVersion = func() string {
+		s := Version
+		if Tag != "" {
+			s += "-" + Tag
+		}
+		return s
+	}()
```
jooola, line 22:
```suggestion
	// version is a semver version (https://semver.org).
	version = "1.2.1" // x-release-please-version

	// versionPrerelease is a semver version pre-release identifier (https://semver.org).
	//
	// For published releases, we set this to an empty string.
	versionPrerelease = "dev"

	// Version of the hcloud CLI.
	Version = func() string {
		s := version
		if versionPrerelease != "" {
			s += "-" + versionPrerelease
		}
		return s
	}()
```

jooola (approved):

PR #656, opened 2024-01-02
feat: add global --quiet flag to hide non-error messages

### LINKED ISSUE #644: json output format should NOT output progress
`hcloud server request-console <server> -o json` now outputs not only json, but also the progress on the STDERR channel:

```
$ hcloud server request-console <server> -o json
1.1s [===================================] 100.00%
{
  "wss_url": "wss://web-console.hetzner.cloud/?server_id=<server_id>\u0026token=<token>",
  "password": "<password>"
}
```

If json is requested, only json should be in the output. Maybe add an optional `--progress`, if that feature should still be present.

Issue was introduced with hcloud v1.41.
last known good is hcloud v1.40.

pauhull:
Closes #644

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -63,6 +63,7 @@ func NewRootCommand(s state.State) *cobra.Command {
 		primaryip.NewCommand(s),
 	)
 	cmd.PersistentFlags().Duration("poll-interval", 500*time.Millisecond, "Interval at which to poll information, for example action progress")
+	cmd.PersistentFlags().Bool("quiet", false, "Only print error messages")
 	cmd.SetOut(os.Stdout)
 	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
 		pollInterval, err := cmd.Flags().GetDuration("poll-interval")
@@ -70,6 +71,14 @@ func NewRootCommand(s state.State) *cobra.Command {
 			return err
 		}
 		s.Client().WithOpts(hcloud.WithPollBackoffFunc(hcloud.ConstantBackoff(pollInterval)))
+
+		if quiet, _ := cmd.Flags().GetBool("quiet"); quiet {
+			f, err := os.Open(os.DevNull)
+			if err != nil {
+				return err
+			}
+			cmd.SetOut(f)
+		}
 		return nil
 	}
 	return cmd
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -64,7 +64,7 @@ func NewRootCommand(s state.State) *cobra.Command {
 	)
 	cmd.PersistentFlags().Duration("poll-interval", 500*time.Millisecond, "Interval at which to poll information, for example action progress")
 	cmd.PersistentFlags().Bool("quiet", false, "Only print error messages")
-	cmd.SetOut(os.Stdout)
+
 	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
 		pollInterval, err := cmd.Flags().GetDuration("poll-interval")
 		if err != nil {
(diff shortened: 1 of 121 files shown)


internal/state/helpers.go
```diff
@@ -131,6 +135,10 @@ func DisplayProgressCircle(cmd *cobra.Command, errCh <-chan error, waitingFor st
 		ellipsis = " ... "
 	)
 
+	if quiet, _ := cmd.Flags().GetBool("quiet"); quiet {
+		return <-errCh
```
jooola, line 139:
```suggestion
		return nil
```

pauhull: Why should we not wait for the action to finish?

jooola: Woups, you are right.

jooola (approved):

internal/cli/root.go
```diff
@@ -62,6 +62,17 @@ func NewRootCommand(s state.State) *cobra.Command {
 		primaryip.NewCommand(s),
 	)
 	cmd.PersistentFlags().Duration("poll-interval", 500*time.Millisecond, "Interval at which to poll information, for example action progress")
+	cmd.PersistentFlags().Bool("quiet", false, "Only print error messages")
 	cmd.SetOut(os.Stdout)
```
jooola, line 66: Could you move this inside to function below, so the logic is self-contained.

internal/cmd/base/create_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/hetznercloud/cli/internal/state"
+	"github.com/hetznercloud/cli/internal/testutil"
+)
+
+type fakeResource struct {
+	ID   int    `json:"id"`
+	Name string `json:"name"`
+}
+
+var commandCalled bool
```
jooola, line 22: Is this really needed, can't you check the returned value of the cmd?

internal/cmd/base/create_test.go
```diff
@@ -0,0 +1,155 @@
...
+	cmd := cli.NewRootCommand(fx.State())
+	fx.ExpectEnsureToken()
+
+	cmd.AddCommand(fakeCreateCmd.CobraCommand(fx.State()))
+
+	out, errOut, err := fx.Run(cmd, []string{"create", "-o=yaml"})
+
+	assert.Equal(t, true, commandCalled)
+	assert.NoError(t, err)
+	assert.YAMLEq(t, `{"resource": {"id": 123, "name": "test"}}`, out)
```
jooola, line 96: That's JSON or I am missing the trick ?

internal/cli/root.go
pauhull: Done in 32afd2fbca6c1b0358a58e5d2c56601b27a11a39

jooola (changes requested):
Could you refactor all the test to use a test helper and use a list of test cases ?

internal/cmd/base/create_test.go
pauhull: Sometimes the output is empty (for example in `TestCreateQuiet`)

internal/cmd/base/create_test.go
pauhull: `assert.YAMLEq` only checks for semantical equivalence so you can use JSON here. I did that here because it's more compact

jooola:
Is there a chance that we fix the cyclic import by moving this to the main.go:

https://github.com/hetznercloud/cli/blob/2df5e92c427dc5770f1d96ff340ad2382f8734f3/internal/cli/root.go#L42-L63

And maybe move the root command somewhere closer to the base package ?

https://github.com/hetznercloud/cli/blob/2df5e92c427dc5770f1d96ff340ad2382f8734f3/internal/cli/root.go#L32

pauhull:
I think we could, but it makes most sense to me to define subcommands where the root command is defined. Also, testing in packages ending with `_test` is good practice anyway (some linters even enforce it) and we were not consistently doing it before anyway. We could move that into another PR though so that the diff is smaller.

jooola:
> I think we could, but it makes most sense to me to define subcommands where the root command is defined. 

In terms of logic, I understand why, though I don't think it will make it less understandable. But in terms of code structure, I'd rather have a tree/diamond shape than those cyclic imports.

Happy to merge this, though.

jooola (approved):

PR #657, opened 2024-01-02
test: increase test coverage

phm07:


internal/cmd/network/disable_protection_test.go
```diff
@@ -0,0 +1,42 @@
+package network_test
+
+import (
+	_ "embed"
```
jooola, line 4: I see that the embed lib is sometimes imported but not used in the file, is that normal?

apricote: That is required for `//go:embed` to work: https://pkg.go.dev/embed

jooola: I understand, but we don't embed any file in this particular file.

apricote: Removed all embed imports in files without actual embeds.

jooola (approved):

PR #659, opened 2024-01-03
refactor: make state mockable

pauhull:


internal/state/state.go
```diff
@@ -2,42 +2,65 @@ package state
...
-type State struct {
+type State interface {
+	TokenEnsurer
+	ActionWaiter
+	context.Context
+	hcapi2.Client
+
+	WriteConfig() error
+	Config() *Config
+}
```
jooola, line 25:
```suggestion
type State interface {
	context.Context

	TokenEnsurer
	ActionWaiter

	Client() hcapi2.Client
	Config() *Config
	// WriteConfig() error => move this to the Config struct e.g. Config().Write(args), not the state responsability
}
```

After looking at the changes, I think it is a good idea if we add a bit of name spacing here. The config and client APIs are quite verbose, while they can be confined and stand on their own.

Our state looks a lot like an extended context, so I moved the context to the top.

And I think it's alright if we leave the command helpers without namespace, I like the idea of injecting other interfaces in the state. We can implement functions that require e.g. the `TokenEnsurer` interface, and still be able to pass the entire State.

internal/state/state.go
```diff
@@ -2,42 +2,65 @@ package state
...
 	if p := os.Getenv("HCLOUD_CONFIG"); p != "" {
 		s.ConfigPath = p
 	}
-	return s
+	if err := s.readConfig(); err != nil {
+		return nil, fmt.Errorf("unable to read config file %q: %s\n", s.ConfigPath, err)
+	}
+	s.readEnv()
+	s.hcloudClient = s.newClient()
+	s.Client = hcapi2.NewClient(s.hcloudClient)
```
jooola, line 55: Great that we grouped the config loading together in the same place, felt a bit all over the place.

But I think we should refrain to have a huge New function, we should probably keep it split into multiple Public functions and run them in the main function.

This might help us with the poll interval flag, as we can delay the config loading and the client creation only after during the root command execution.

internal/cli/root.go
```diff
@@ -41,26 +40,26 @@ func NewRootCommand(state *state.State, client hcapi2.Client) *cobra.Command {
 		DisableFlagsInUseLine: true,
 	}
```
jooola, line 41: For example, we could load the config/client here using the `PersistentPreRun`.

internal/state/state.go
pauhull: Implemented in 6597bd5fe7c10b10bbbe860795530db30e8626b0

internal/cli/root.go
pauhull: Not possible because fully configured state is needed for root command initialization -> Out of scope for this PR

internal/state/helpers.go
```diff
@@ -20,51 +18,23 @@ const (
 	progressBarTpl    = `{{ etime . }} {{ bar . "" "=" }} {{ percent . }}`
 )
 
-func (c *State) Wrap(f func(*State, *cobra.Command, []string) error) func(*cobra.Command, []string) error {
+func Wrap(s State, f func(State, *cobra.Command, []string) error) func(*cobra.Command, []string) error {
```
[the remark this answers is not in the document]
pauhull: Because it would need to be reimplemented on each implementation of `state.State` even though it is a universal util function. The change was necessary because you can't define functions on interfaces.

jooola (approved):
:fire:

PR #665, opened 2024-01-09
feat: allow adding/removing multiple labels at once

### LINKED ISSUE #662: Allow adding/removing multiple labels at once
Currently, the label commands only allow a single label to be passed as argument.

This could be improved, and should allow updating multiple labels at once.
```
hcloud server add-label SERVER LABEL... 
hcloud server add-label 12345 key=value foo=bar another=one 
```

pauhull:
Closes #662


jooola:
The `validateAddLabel` and `validateRemoveLabel` functions might need updating too.

pauhull:
I updated `validateAddLabel` to account for the new changes

apricote (approved):
Tested locally, works great :)

PR #671, opened 2024-01-10
feat: allow setting default SSH keys for contexts

pauhull:


internal/cmd/context/ssh_key.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	cmd.AddCommand(newSSHKeyAddCommand(s))
+	cmd.AddCommand(newSSHKeyRemoveCommand(s))
+	cmd.AddCommand(newSSHKeyListCommand(s))
+	return cmd
+}
+
+func newSSHKeyAddCommand(s state.State) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "add SSH-KEY...",
```
apricote, line 36: I feel this breaks the behaviour of `hcloud context`. All other subcommands take the name of the context as a positional argument, while this one takes the ssh-key as positional and context as flag.

internal/cmd/server/create.go
```diff
@@ -261,6 +261,11 @@ func createOptsFromFlags(
 	primaryIPv6IDorName, _ := flags.GetString("primary-ipv6")
 	protection, _ := flags.GetStringSlice("enable-protection")
 
+	// check if ssh-key flag was set, otherwise set to defaults
+	if ctx := s.Config().ActiveContext; ctx != nil && !flags.Changed("ssh-key") {
+		sshKeys = ctx.SSHKeys
+	}
```
apricote, line 267: I think we should add this to the help output for this command. Same for `enable-rescue`.

What do you think about logging an info text that the default keys were used to create the token? Otherwise the customer might have forgotten about it and expects an email with the password.

apricote (changes requested):

internal/cmd/server/create.go
pauhull: Yes, that sounds useful. I'll add that.

internal/cmd/context/ssh_key.go
apricote: You can still make it an optional positional arg, defaulting to the current index.

I checked some other commands and they all have the pattern `hcloud <resource> <action> <resource-name-or-id>`. The commands matching our current pattern would be:

- `hcloud context add-ssh-keys [context] --ssh-key foobar`
- `hcloud context remove-ssh-keys [context] --ssh-key foobar`
- `hcloud context list-ssh-keys [context] --ssh-key foobar`

I just noticed that you actually added another level to the command, so the pattern mentioned above actually matches the last three segments. And I am kinda fine with it.

@jooola any opinions?

internal/cmd/context/ssh_key.go
```diff
@@ -0,0 +1,211 @@
...
+		Short:                 "Add a default SSH key to the context",
+		TraverseChildren:      true,
+		DisableFlagsInUseLine: true,
+		RunE:                  state.Wrap(s, runSSHKeyAdd),
+	}
+	cmd.Flags().String("context", "", "Name of the context to add the default SSH key(s) to")
+	_ = cmd.RegisterFlagCompletionFunc("context", cmpl.SuggestCandidates(config.ContextNames(s.Config())...))
```
apricote, line 41: What do you think about moving this flag to `newSSHKeyCommand`, as its specified in every subcommand?

internal/cmd/context/ssh_key.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	cmd.AddCommand(newSSHKeyAddCommand(s))
+	cmd.AddCommand(newSSHKeyRemoveCommand(s))
+	cmd.AddCommand(newSSHKeyListCommand(s))
+	return cmd
+}
+
+func newSSHKeyAddCommand(s state.State) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "add SSH-KEY...",
```
apricote, line 34: Could you add autocompletion for the available SSH Keys to this command?

internal/cmd/context/ssh_key.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if err := s.Config().Write(); err != nil {
+		return err
+	}
+
+	cmd.Printf("Added SSH key(s) %s to context \"%s\"\n", strings.Join(keys, ", "), ctx.Name)
+	return nil
+}
+
+func newSSHKeyRemoveCommand(s state.State) *cobra.Command {
```
apricote, line 112: Could you add autocomplete for `SSH-KEY` to this command?

internal/cmd/context/ssh_key.go
pauhull: Good idea. What do you think a good description would be?

internal/cmd/context/ssh_key.go
pauhull: Done in d87cdc1

internal/cmd/context/ssh_key.go
jooola: No strong opinion, but I prefer less nested commands, discovery is easier when one need to read the help pages.

About the flags, I would follow the convention we already use in the CLI.

Using `hcloud context remove-ssh-keys [context] --ssh-key foobar`

pauhull: I think having a dedicated subcommand helps with sorting and provides a better structure.
With the proposed subcommands the help would look like this:
```
$ hcloud context -h
Manage contexts

Usage:
  hcloud context [command]

Available Commands:
  active      Show active context
  add-ssh-keys ...
  create      Create a new context
  delete      Delete a context
  list        List contexts
  list-ssh-keys ...
  remove-ssh-keys ...
  use         Use a context
```

jooola: Is it possible to order the command manually?

apricote: We can use cobra `Groups` to group the output into various ... groups. I would love to see that anyway for `hcloud server` and other huge commands.

pauhull: I still think having a dedicated subcommand for it is nice because command names are kept shorter and the `hcloud context ssh-key` command follows the rough structure of other subcommands (add/remove/list).
Yes, I agree that we should group subcommands for larger commands.

PR #672, opened 2024-01-10
test: make state config mockable

pauhull:


internal/state/config/config.go
```diff
@@ -0,0 +1,177 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	toml "github.com/pelletier/go-toml/v2"
+)
+
+type Config interface {
+	Write() error
+	Marshal() ([]byte, error)
```
apricote, line 13: This is only called from `Config.Write()` and can be removed from the interface.

jooola (approved):

internal/hcapi2/mock/mock_gen.go
```diff
@@ -1,5 +1,6 @@
 package hcapi2_mock
 
+//go:generate mockgen -package hcapi2_mock -destination zz_config_mock.go github.com/hetznercloud/cli/internal/state/config Config
```
apricote, line 3: I dont think this belongs into `internal/hcapi2/mock`. Its not a mock for `hcapi`. Perhaps `internal/state/config/mock`?

pauhull: Yes, I already realized that :D

internal/state/config/config.go
pauhull: What do you think about removing `ContextNames`, `ContextByName` and `RemoveContext` too? Should they stay mockable or can we just make a utility function that takes a `config.Config` and implements the functionality using `Contexts` and `SetContexts`?

apricote: Fine by me, if it makes the code cleaner

pauhull: Done in 6292032 and 68a7918

apricote (approved):

PR #679, opened 2024-01-18
feat(server): remove unsupported linux32 rescue type

phm07:
The `linux32` rescue type has been removed and is now unsupported ([See docs](https://docs.hetzner.cloud/changelog#2024-01-08-server-action-enable-rescue-no-longer-supports-32bit)). This PR removes it from suggestions and also checks for the passed type to be valid before sending an API request (like how it's done with other enums).

### DIFF
--- a/internal/cmd/server/enable_rescue.go
+++ b/internal/cmd/server/enable_rescue.go
@@ -23,7 +23,7 @@ var EnableRescueCmd = base.Cmd{
 			DisableFlagsInUseLine: true,
 		}
 		cmd.Flags().String("type", "linux64", "Rescue type")
-		cmd.RegisterFlagCompletionFunc("type", cmpl.SuggestCandidates("linux64", "linux32"))
+		cmd.RegisterFlagCompletionFunc("type", cmpl.SuggestCandidates("linux64"))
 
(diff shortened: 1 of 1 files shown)


apricote (approved):

internal/cmd/server/enable_rescue.go
```diff
@@ -43,7 +43,14 @@ var EnableRescueCmd = base.Cmd{
 			opts hcloud.ServerEnableRescueOpts
 		)
 		rescueType, _ := cmd.Flags().GetString("type")
+
 		opts.Type = hcloud.ServerRescueType(rescueType)
+		switch opts.Type {
+		case hcloud.ServerRescueTypeLinux64:
+			break
+		default:
+			return fmt.Errorf("invalid rescue type: %v", opts.Type)
```
jooola, line 52: Could we have 1 extra case: 1 for unknown values, and 1 for deprecated values: the error message feels like a user error, but we dropped support for the rescue type.

phm07: Implemented in 4de80c2

jooola (approved):

PR #690, opened 2024-02-06
docs: add available types to hcloud image list --help

kranurag7:
As of now, I had to use `-oyaml | yq` to filter out the supported types.  
It would be good to have it in the help message itself.


apricote (changes requested):
Very nice! The `--help` output is pretty useless in many cases and the API Docs are often required to figure out what each flag does. Updating these is on our roadmap for the coming months.

Do you have the shell completion installed? It would also show you the possible options.

apricote (approved):

PR #696, opened 2024-02-12
fix: removing last rule from firewall fails with invalid_input error

kossmac:
If you try to delete last firewall rule, rule is `null` but should be `[]`, resulting in error message 

> hcloud: invalid input in field 'rules' (invalid_input)

`hcloud firewall delete-rule --direction in --protocol tcp --source-ips '8.8.8.8/32' --port 443 f2b`


apricote (approved):
Awesome!

Updated the title for our Changelog.

PR #707, opened 2024-03-05
feat: add update check

phm07:
This PR adds an update checker that checks for a new GitHub release every 24 hours and compares it to the current version.


cmd/hcloud/main.go
```diff
@@ -47,6 +49,8 @@ func main() {
 		log.Fatalf("unable to read config file %q: %s\n", configPath, err)
 	}
 
+	update.CheckForUpdate(cfg)
+
```
jooola, line 53: I don't think this is a good idea. I'd prefer a command for manual check over an automated check.

If we were to implement this automation, I'd make is Opt-In instead of Opt-Out. The users might not want to call GitHub every 24h, for multiple reasons.

This might also become a problem for packages in distributions, where the CLI cannot be updated outside the upgrade policy of the distribution.

phm07: I think an update checker is only effective when it nags the user. I doubt most users would manually check for updates, especially when they don't use the CLI that often.
You're right about updates potentially not being available for your distribution, but that is more an issue of where we get the latest version information from. Maybe we could also integrate some check to disable it altogether because distributions often do update checks themselves.

apricote: ## Automated vs Manual

I prefer an automated check here, people who would do manual checks can also do it through GitHub or their package manager.

## Privacy & Consent

I agree that it is not very nice to call GitHub regularly without users consent. We could make it opt-in and keep calling GitHub. Alternatively we could setup a proxy in our infrastructure that provides the same data, but without sending user IPs to GitHub. As the CLI is calling our API anyway, I do not think there are any privacy issues with this approach and we could make it opt-out.

## Distributions

For users who install the cli from their distribution, showing this check does not make any sense, as they would update it through `apt`, `pacman`, `yum`..., we should provide some way for a packager to disable these checks.

This could be implemented by providing a variable `var UpdateCheckDefault = "true"` that packager can override during the build by setting `go build .. -ldflags '-X internal/version.UpdateCheckDefault=false`.

Another option would be build constraints, where we have a no-op implementation in addition to the current one. The packager could then decide between the two by running `go build -tags noupdatecheck` which would disable the actual implementation and use no-op instead.

apricote (changes requested):

apricote:
After some consideration we want to write a Design Docs before continuing with this PR.

PR #709, opened 2024-03-07
docs: improve command usage documentation

### LINKED ISSUE #708: Improve commands help messages
### TL;DR

The following two --help outputs have inconsistent usage description:
```
Usage:
  hcloud server attach-to-network [FLAGS] SERVER

Usage:
  hcloud network add-subnet NETWORK FLAGS
```

Order of flags and arguments should not matter, but for consistency, we should always propose the same format in the help output.

Another details, is that sometimes we use `[FLAGS]` with brackets marking it as optional, and sometimes not. But according to the further output of `hcloud server attach-to-network --help` the flags here are not optional either, it lists `--network` as a required flag.

We might also want to stop yelling at our users, and not use UPPERCASE keywords unless really necessary.

### Expected behavior

We could use the http://docopt.org/ flags and arguments description language to follow a well known convention.

Related to #700

phm07:
This PR makes command usage documentation consistent by enforcing [docopt](http://docopt.org/) rules and also defining and documenting other additional conventions.

Fixes #708

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,8 @@ Command usage descriptions (namely `cobra.Command.Use`) should follow the [docop
 Additionally:
 - Optional flags should be the first in the usage string.
   - If there is more than one optional flag, you can use `[options]` to abbreviate.
-- Required flags should always be documented explicitly and after the positional arguments.
+- Required flags should always be documented explicitly and before the positional arguments, but after the 
+  optional flags.
 
 ### Subcommand groups
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,15 @@ goreleaser --snapshot --skip-publish --rm-dist
 
 ## Conventions
 
+### Command usage
+
+Command usage descriptions (namely `cobra.Command.Use`) should follow the [docopt](http://docopt.org/) syntax.
+
+Additionally:
+- Optional flags should be the first in the usage string.
+  - If there is more than one optional flag, you can use `[options]` to abbreviate.
+- Required flags should always be documented explicitly and after the positional arguments.
+
 ### Subcommand groups
 
 Cobra offers the functionality to group subcommands. The conventions on when and how to group commands are as follows:
--- a/internal/cmd/certificate/create.go
+++ b/internal/cmd/certificate/create.go
@@ -17,7 +17,7 @@ import (
 var CreateCmd = base.CreateCmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
-			Use:   "create [options] --name <name> (--type managed | --type uploaded --cert-file <file> --key-file <file>)",
+			Use:   "create [options] --name <name> (--type managed --domain <domain> | --type uploaded --cert-file <file> --key-file <file>)",
 			Short: "Create or upload a Certificate",
 			Args:  cobra.ExactArgs(0),
 		}
--- a/internal/cmd/certificate/create.go
+++ b/internal/cmd/certificate/create.go
@@ -17,7 +17,7 @@ import (
 var CreateCmd = base.CreateCmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
-			Use:   "create [FLAGS]",
+			Use:   "create [options] --name <name> (--type managed | --type uploaded --cert-file <file> --key-file <file>)",
 			Short: "Create or upload a Certificate",
 			Args:  cobra.ExactArgs(0),
 		}
--- a/internal/cmd/loadbalancer/disable_public_interface.go
+++ b/internal/cmd/loadbalancer/disable_public_interface.go
@@ -14,7 +14,7 @@ import (
 var DisablePublicInterfaceCmd = base.Cmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		return &cobra.Command{
-			Use:                   "disable-public-interface [options] <load-balancer>",
+			Use:                   "disable-public-interface <load-balancer>",
 			Short:                 "Disable the public interface of a Load Balancer",
 			Args:                  cobra.ExactArgs(1),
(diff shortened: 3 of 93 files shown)


CONTRIBUTING.md
```diff
@@ -35,6 +35,15 @@ goreleaser --snapshot --skip-publish --rm-dist
 
 ## Conventions
 
+### Command usage
+
+Command usage descriptions (namely `cobra.Command.Use`) should follow the [docopt](http://docopt.org/) syntax.
+
+Additionally:
+- Optional arguments should be the first in the usage string.
+  - If there is more than one optional argument, you can use `[options]` to abbreviate.
```
jooola, line 44: Why not make our lives easier and always use options for any optional flags? Unless we have an automated way to handle this, I think it's going to be difficult to maintain and error prone.

CONTRIBUTING.md
```diff
@@ -35,6 +35,15 @@ goreleaser --snapshot --skip-publish --rm-dist
 
 ## Conventions
 
+### Command usage
+
+Command usage descriptions (namely `cobra.Command.Use`) should follow the [docopt](http://docopt.org/) syntax.
+
+Additionally:
+- Optional arguments should be the first in the usage string.
```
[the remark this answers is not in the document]
phm07: Sorry, you're right, I meant flags there. Docopt doesn't say anything about argument/flag order. Having optional flags first, then arguments and then required flags made the most sense to me because then the ordering would be most specific to least specific.

internal/cmd/certificate/create.go
```diff
@@ -17,7 +17,7 @@ import (
 var CreateCmd = base.CreateCmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
-			Use:   "create [FLAGS]",
+			Use:   "create [options] --name <name> (--type managed | --type uploaded --cert-file <file> --key-file <file>)",
```
apricote, line 20: `--domain` is required for `managed` certificates, see line 117.

internal/cmd/loadbalancer/disable_public_interface.go
```diff
@@ -14,7 +14,7 @@ import (
 var DisablePublicInterfaceCmd = base.Cmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		return &cobra.Command{
-			Use:                   "disable-public-interface [FLAGS] LOADBALANCER",
+			Use:                   "disable-public-interface [options] <load-balancer>",
```
apricote, line 17:
```suggestion
			Use:                   "disable-public-interface <load-balancer>",
```

There are no other flags.

internal/cmd/server/create_image.go
```diff
@@ -15,7 +15,7 @@ import (
 var CreateImageCmd = base.Cmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
-			Use:   "create-image [FLAGS] SERVER",
+			Use:   "create-image [options] <server>",
```
apricote, line 18:
```suggestion
			Use:   "create-image [options] <server> --type <snapshot|backup>",
```

`--type` is required.

internal/cmd/server/metrics.go
```diff
@@ -29,7 +29,7 @@ var metricTypeStrings = []string{
 var MetricsCmd = base.Cmd{
 	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
-			Use:                   "metrics [FLAGS] SERVER",
+			Use:                   "metrics [options] <server>",
```
apricote, line 32: `--type` is required.

ow are required flags with enum values that can be specified multiple times defined in docopt? `(--type <cpu,disk,network>)...` perhaps?

apricote (changes requested):
Very nice!

internal/cmd/server/metrics.go
phm07: I don't think there is a standardized approach to this. I like your suggestion. Maybe we can use the regular convention for enums inside the angled brackets (`cpu|disk|network`) to make it clear that it's an enum. Then `...` would still work: `(--type <cpu|disk|network>)...` <-> `[(--type cpu) (--type disk) (--type network)]`

phm07: Yes, I meant to point out that this is what the syntax would expand to ("proving" that it is still valid). Of course we wouldn't want this for the actual usage string.

CONTRIBUTING.md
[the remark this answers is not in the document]
phm07: About the ordering, this is what the CLI Guidelines have to say: https://clig.dev/#help
They suggest putting the `[options]` first as well (like jq and git do it too for example). But generally they put all flags first (even required ones) and the positional arguments last, while we do optional flags < positional args < required flags. Maybe we can move the positional args all the way to the back like jq does. 
On the other hand, if you look [here](https://github.com/hetznercloud/cli/pull/709/files#diff-fd9e246bf99bd641dacfa0eb236373abe4720c1398ad0a468266b3357c92b845R20) for example, positional args would be easily overlooked if they are placed last. What's your opinion on this?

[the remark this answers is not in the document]
jooola: I would always try to have the required and optional flags next to each other, having the arguments in the middle seem weird.

Whether we put the arguments at the beginning or the end, I think I prefer at the beginning. When I write a command, I prefer to go in order of importance `<resource type> <actions> <resource id> <resource fields (e.g. labels)>`. Having the flags before also seem weird.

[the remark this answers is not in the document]
apricote: I do not have a strong opinion on this.

While using the CLI I use the same mental model as @jooola and also put `<args> <req flags> <optional flags>`.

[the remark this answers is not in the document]
apricote: It looks like you only committed some of the changes, there are still many files with old order, for example `internal/cmd/loadbalancer/update_service.go` or `internal/cmd/firewall/remove_from_resource.go`

apricote (approved):

PR #711, opened 2024-03-08
feat: improve missing argument error messages

### LINKED ISSUE #700: Improve error message when required argument is not set
### TL;DR

If a user does not pass a required (positional) argument, we return the generic error message `hcloud: accepts 1 arg(s), received 0`. This does not help the user in figuring out what the argument should be.

### Expected behavior

I would prefer to see something along the lines of

```
hcloud: No $RESOURCE name was specified, please add the argument. To learn more, add the flag --help.
```

Right now we use the default `cobra` argument validators (mostly `cobra.ExactArgs(1)`) which produce this less-than-useful errors. We will need to write our own version of these validators with improved messages and maybe open an issue upstream to get better versions in cobra.

phm07:
This PR adds a custom Cobra command validator that outputs useful errors when positional arguments are missing during command execution. For example, such an error could look like this:
```
$ hcloud server describe   
describe [options] <server>
                    ^^^^^^
hcloud: expected argument server at position 1
```

Where previously it would look like this:
```
$ hcloud server describe
hcloud: accepts 1 arg(s), received 0
```

This is done by parsing the `Use` property of a Cobra command with a regular expression and then matching the corresponding missing arguments.

Fixes #700


internal/cmd/util/validation.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var posArgRegex = regexp.MustCompile(`\[.*?]|\(.*?\)|--.*?<.*?>|<(.*?)>`) // matches all positional args in group 1
```
jooola, line 12: I find the use of a regex for such feature a bit fragile, why didn't you choose to have a list of args and flags that could be both rendered to a usage string and used to validate the missing args ?

apricote: There is also https://github.com/docopt/docopt.go that might be used to get the info we want.

phm07: Rendering args and flags to a usage string could be complicated in some cases, for example [here](https://github.com/hetznercloud/cli/pull/711/files#diff-37acc61009ac872d74e2bbfe664618f99e6ba1e9497c01dd65f1fd3f4a5553b7R19) and [here](https://github.com/hetznercloud/cli/pull/711/files#diff-09ce77c4101f651b597cdb7b6f14a4252ad407a7b0999abe9bf6cc8419ffdd5aR19). Parsing seems like the best bet. Maybe we can add a comment to explain the regex.

internal/cmd/util/validation.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+	if len(args) > len(expected) && exact {
+		return fmt.Errorf("expected exactly %d positional arguments, but got %d", len(expected), len(args))
+	}
+	for i := 0; i < len(expected); i++ {
+		if i >= len(args) {
+			idx := strings.Index(cmd.Use, "<"+expected[i]+">")
+			if idx != -1 {
+				_, _ = fmt.Fprintln(os.Stderr, cmd.Use)
```
apricote, line 37: I love the helpful error messages here! I have one suggestion, but if its too hard to implement or you would prefer to do it in a follow up PR that is fine by me :)

Right now this prints something like:

```shell
➜  hcloud server delete
delete <server>
        ^^^^^^
hcloud: expected argument server at position 1
```

I think the output could be a bit clearer if it would include all parent commands, not only the current command we are validating:

```shell
➜  hcloud server delete
hcloud server delete <server>
                      ^^^^^^
hcloud: expected argument server at position 1
```

internal/cmd/util/validation_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Use: "test <arg1> <arg2>",
+		}, []string{"arg1", "arg2", "arg3"})
+	})
+
+	assert.Empty(t, stdout)
+	assert.Empty(t, stderr)
+	assert.NoError(t, err)
+}
+
+func TestValidateExactComplexUsage(t *testing.T) {
```
apricote, line 60: To verify that the regex works correctly, could you add "Complex Usage" tests that have too few and too many arguments?

internal/cmd/util/validation_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/hetznercloud/cli/internal/testutil"
+)
+
+func TestValidateExact(t *testing.T) {
```
apricote, line 12: As the structure of all these tests is exactly the same, you could use [Table Driven Tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) to simplify adding new test cases.

internal/cmd/base/delete.go
```diff
@@ -34,7 +34,7 @@ func (dc *DeleteCmd) CobraCommand(s state.State) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:                   fmt.Sprintf("delete %s<%s>", opts, strings.ToLower(dc.ResourceNameSingular)),
 		Short:                 dc.ShortDescription,
-		Args:                  cobra.ExactArgs(1),
+		Args:                  util.ValidateExact,
```
apricote, line 37: The Base Command already sets a default for every command that does not define `Args`: https://github.com/hetznercloud/cli/blob/b9265f42ad57d8f35e28c30ef7e3e45cb2341753/internal/cmd/base/cmd.go#L21-L23

You could update this default to `util.ValidateExact` and remove `Args: util.ValidateExact` from every command that uses `base.Cmd` directly or through `base.ListCmd`, `base.CreateCmd` (...).

A potential follow up could be to make sure that every command uses `base.Cmd`, at a quick glance it was mostly the no-op parent commands that made no use of it so far. perhaps it makes sense to introduce a no-op `base.ParentCmd` that only hosts children. That way we could also remove some other defaults set in `base.Cmd` from all actual commands: `TraverseChildren` and `DisableFlagsInUseLine`.

internal/cmd/util/validation.go
```diff
@@ -0,0 +1,44 @@
...
+
+var posArgRegex = regexp.MustCompile(`\[.*?]|\(.*?\)|--.*?<.*?>|<(.*?)>`) // matches all positional args in group 1
+
+func ValidateExact(cmd *cobra.Command, args []string) error {
+	return validate(cmd, args, true)
+}
+
+func Validate(cmd *cobra.Command, args []string) error {
+	return validate(cmd, args, false)
+}
```
apricote, line 20: Could you add some doc strings to these functions to explain what they do and how they differ?

In addition it would be nice to have this in the `CONTRIBUTING.md`.

apricote (changes requested):

internal/cmd/util/validation.go
apricote: Yea, I was suprised about the `hcloud: %w` error output.

apricote (approved):
Looks good to me :) Can be merged once #709 is in `main` :)

PR #712, opened 2024-03-12
test: increase test coverage

phm07:


internal/cmd/util/deprecation_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (t testDeprecatable) UnavailableAfter() time.Time {
+	return t.unavailableAfter
+}
+
+func (t testDeprecatable) DeprecationAnnounced() time.Time {
+	return t.deprecationAnnounced
+}
+
+func TestDescribeDeprecation(t *testing.T) {
```
apricote, line 32: What do you think about adding another test (or convert to table test) for a not-deprecated resource? That would fully cover the function under test.

apricote (approved):

apricote (approved):

PR #713, opened 2024-03-12
chore: restore mock_gen.go

phm07:
This file was erroneously deleted in #672


phm07:
To prevent this in the future, we should set up a CI pipeline where all generated files are deleted and re-generated and then it is checked that the diff is empty. That would also detect possible changes in generated files.

jooola (approved):

PR #719, opened 2024-03-20
feat: allow deletion of multiple resources at once

phm07:
Currently, the CLI only allows you to delete one resource at a time. This can be time consuming and tedious if you want to delete multiple resources, because you have to wait for each resource to be deleted before you can delete the next one.
This PR adds the ability to delete multiple resources at once by passing them to the `delete` subcommand as multiple positional arguments, e.g. `hcloud server delete server1 server2 ...`. It also adds tests to verify that this works as intended.

### DIFF
--- a/internal/cmd/base/delete_test.go
+++ b/internal/cmd/base/delete_test.go
@@ -41,8 +41,16 @@ func TestDelete(t *testing.T) {
 			Args:   []string{"delete", "123"},
 			ExpOut: "Fetching fake resource\nDeleting fake resource\nFake resource 123 deleted\n",
 		},
+		"no flags multiple": {
+			Args: []string{"delete", "123", "456", "789"},
+			ExpOut: "Fetching fake resource\nDeleting fake resource\nFake resource 123 deleted\nFetching fake resource\n" +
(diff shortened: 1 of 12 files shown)


internal/cmd/base/delete.go
```diff
@@ -27,9 +28,9 @@ type DeleteCmd struct {
 // CobraCommand creates a command that can be registered with cobra.
 func (dc *DeleteCmd) CobraCommand(s state.State) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                   fmt.Sprintf("delete [FLAGS] %s", strings.ToUpper(dc.ResourceNameSingular)),
+		Use:                   fmt.Sprintf("delete [FLAGS] %s...", strings.ToUpper(dc.ResourceNameSingular)),
 		Short:                 dc.ShortDescription,
-		Args:                  cobra.ExactArgs(1),
+		Args:                  cobra.MinimumNArgs(1),
 		ValidArgsFunction:     cmpl.SuggestArgs(cmpl.SuggestCandidatesF(dc.NameSuggestions(s.Client()))),
```
apricote, line 34: I installed this PR locally to verify autocomplete, but noticed that it only autocompletes the first entry. For subsequent autocompletes I only get suggestions for local files. It would be nice to continue to get proper suggestions.

apricote (approved):

PR #729, opened 2024-04-11
fix: allow passing multiple args to server ssh command

### LINKED ISSUE #728: hcloud server ssh <server> [command] is broken
### TL;DR

executing a remote command over ssh is no longer possible

### Expected behavior

when running `hcloud server ssh <server> [command]` I expect the command to be run on the remote server.

### Observed behavior

I get this error in the terminal:
hcloud server ssh [options] <server> [command]
                                               ^
hcloud: expected exactly 1 positional argument(s), but got 2

### Minimal working example

download v1.43.0
setup context and have a server ready, then:
`hcloud server ssh demoserver "ls"`

### Log output

```plain text
hcloud server ssh [options] <server> [command]
                                               ^
hcloud: expected exactly 1 positional argument(s), but got 2
```

### Additional information

it worked fine in previous version 1.42.0

jooola:
Related to #711

Closes #728

### DIFF
--- a/internal/state/config/config.go
+++ b/internal/state/config/config.go
@@ -19,6 +19,8 @@ type Config interface {
 	SetContexts([]*Context)
 	Endpoint() string
 	SetEndpoint(string)
+
+	SSHPath() string
 }
 
 type Context struct {
@@ -92,6 +94,10 @@ func (cfg *config) Endpoint() string {
 
 func (cfg *config) SetEndpoint(endpoint string) {
 	cfg.endpoint = endpoint
+}
+
+func (cfg *config) SSHPath() string {
+	return "ssh"
 }
 
 func ContextNames(cfg Config) []string {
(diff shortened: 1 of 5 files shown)


apricote:
Can we add some test cases for this command? It is pretty special in that it just spawns `ssh`, so I would like to make sure that it will keep working in the future.

jooola:
Not totally happy with how the test are right now, might need some more refactoring.

internal/state/config/config.go
```diff
@@ -94,6 +96,10 @@ func (cfg *config) SetEndpoint(endpoint string) {
 	cfg.endpoint = endpoint
 }
 
+func (cfg *config) SSHPath() string {
+	return "ssh"
+}
```
jooola, line 101: Maybe this could become a real user preference? And not only be a workaround for mocking.

Issue #733, opened 2024-04-18
`--port` option is missing from usage of the `firewall {add,delete}-rule` commands


jooola:
### TL;DR

The `--port` flag is missing from the `hcloud firewall add-rule` and `hcloud firewall delete-rule` commands.

I don't know if this is expected, are only required flags part of the usage?

### Expected behavior

That the `--port` flag is also part of the command usage.

### Observed behavior

```
$ hcloud firewall add-rule --help
Add a single rule to a firewall

Usage:
  hcloud firewall add-rule [options] (--direction in --source-ips <ips> | --direction out --destination-ips <ips>) --protocol <icmp|udp|tcp|esp|gre> <firewall>

Flags:
      --description string            Description of the firewall rule
      --destination-ips stringArray   Destination IPs (CIDR Notation) (required when direction is out)
      --direction string              Direction (in, out) (required)
  -h, --help                          help for add-rule
      --port string                   Port to which traffic will be allowed, only applicable for protocols TCP and UDP, you can specify port ranges, sample: 80-85
      --protocol string               Protocol (icmp, esp, gre, udp or tcp) (required)
      --source-ips stringArray        Source IPs (CIDR Notation) (required when direction is in)
```

### Minimal working example

$ hcloud firewall add-rule --help
$ hcloud firewall delete-rule --help

apricote:
I think this was missed because `--port` is not a globally required flag, only when using `--protocol <tcp|udp>` it becomes required.

---
While testing this, I also found that we do not print the `invalid_input` details. Which would help users in case our local validation is missing something:

```
--- Response:
{
  "error": {
    "message": "invalid input in field 'rules'",
    "code": "invalid_input",
    "details": {
      "fields": [
        {
          "name": "rules.port",
          "messages": [
            "tcp and udp protocols requires the port parameter to be specified"
          ]
        }
      ]
    }
  }
}

--- CLI Output:
hcloud: invalid input in field 'rules' (invalid_input)
```

apricote:
Fixed by #734 and #735.

PR #735, opened 2024-04-19
docs(firewall): port flag in add/delete-rule usage string

phm07:
See #733


jooola (approved):
This change also have an influence on the behavior of the argument validation, correct? Should we scope this as `feat` or `fix` ?

I haven't tested it manually, but the code looks good.

apricote (approved):

Issue #434, opened 2022-12-13
feat: Additional Settings (e.g. sort preference, default columns)


apricote:
We should discuss whether we want to add additional settings to the CLI, such as configuring sort preferences (#418 #421) and the default column selection (#423). If we decide that we want additional settings, we need to find an easy to maintain way to read and modify them.

- Where to save the settings?
- What interface do we need to provide for internal usage?
- Should this be configurable per-context or globally?

apricote:
@cedi I would like wait with your PRs until I had time to discuss this with my colleagues. As its about to be holiday season, this will probably be delayed until January.

apricote:
Only the default columns are missing now. Could you create a new ticket for that request and close this issue @phm07?

PR #736, opened 2024-04-19
feat: new configuration system, config subcommand

### LINKED ISSUE #762: [Design Doc] feat: new configuration system + config subcommand
# Glossary

### Option

An option is a key-value pair that modifies how the CLI behaves.

Examples:

- Default SSH Keys
- Debug Logging Enabled
- Current Active Context
- API Token

### Context

A context is a named list of options that the user can enable to apply all options defined in it. It always includes an API Token. Only a single context can be active at a time.

### Configuration File

The configuration file contains a collection of contexts and global options. It is being read by the CLI.

### Preferences

The subset of options that are not mandatory to configure before the CLI can be used.

### Configuration

All the ways the user can set options for the CLI. Including the Configuration File, Environment Variables and Flags.

# Motivation

Our CLI already uses a configuration file to save the available `contexts` (API Tokens + Name) as well the current selected context. Modifying the file is done through sub-commands `hcloud context create|delete|use|active|list`

- The CLI also has some global options exposed through environment variables
    - `HCLOUD_DEBUG`
    - `HCLOUD_TOKEN`
    - `HCLOUD_CONFIG`
    - `HCLOUD_ENDPOINT`
    - `HCLOUD_DEBUG_FILE`
    - `HCLOUD_CONTEXT`
- The CLI also has some global options exposed through flags
    - `--poll-interval`
    - `--quiet`

The config is saved to a file in (UNIX) `~/.config/hcloud/cli.toml`.

The current options are badly/not documented. It is not clear where you can set which options.

- We are currently thinking about adding further config options
    - Default SSH keys
    - Table sort
    - Table columns
    - Default location
    - Update Checks

This Design Doc is intended to find a good solution that can handle the current and future options while keeping ways of configuration consistent.

# Implementation

### What are Options?

- Key/Value pairs
- Keys are always `strings`  in kebab-case.
- Keys can be namespaced with `.`
- Values can have these data types for the :
    - `string`, `integer`, `float`, `bool`
    - Any of the above as `array`

### UX / How does the user interact with the configuration file?

We will introduce a new `hcloud config` sub-command that provides a generic interface to modify options (from the preferences) in the configuration file.

Usages:

- `hcloud config --help`
    - Explains all options and their default values
- `hcloud config [--global] get <key>`
    - Shows the current value of key
    - In the future: show origin of value with `--show-origin` flag (config, env, flags…)
- `hcloud config [--global] set <key> <value>...`
    - Set/Replace the value of a key
- `hcloud config [--global] unset <key>`
    - Delete an option from the config
- `hcloud config [--global] <add|remove> <key> <value>...`
    - Only applies to `array` options. Adds or removes the values from the key.
- `hcloud config [--global] list`
    - Lists all configured options

We plan to add autocomplete for option keys. For option values we want to add autocomplete if possible. This can be done with Cobra’s [`ValidArgsFunction`](https://github.com/spf13/cobra/blob/a0a6ae020bb3899ff0276067863e50523f897370/command.go#L87).

By default, this command modifies the options saved in the current active context. Changing this can be done through the `context` option (`--context`, `HCLOUD_CONTEXT`). To change options on the global scope, the user can specify the `--global` flag.

Using `--global` as opt-in, we prevent accidental options overwrites. 

### How to deal with existing flags?

If the option defines a default that already has one or multiple flags (like `default-ssh-keys` in `hcloud server create [--ssh-key]`) we do not add new flags for these options. The existing flag would default to the option if the option is set. If the actual flag was set, the default option is overridden.

Options that are not “defaults for existing flags” should get their own flag at the root command level. 

For other global existing Environment Variables and Flags we will try to unify them into the new configuration pattern.

### How do we structure the config?

We introduce a new `preferences` object to the configuration file schema. This holds all optional options (see below). It can be specified on the top level for global preferences and inside every context for context-specific preferences.

The only options not included in the `preferences` are:

- (Global) `active-context`
- (Context) `name`
- (Context) `token`

Namespaced option keys are set in a nested object inside of `preferences`.

Preferences configured globally apply to all contexts. If a context sets a specific option in the context preferences, this override the global value.

Users can “unset / ignore” global preferences on a context-level by assigning the zero value for its data-type (`””`, `0`, `0.0`). Unsetting booleans is not possible. This allows to retreive the next  available option value by precedence, skipping the global preferences.

The keys used in the configuration file equal the option keys, but we replace any `-` with `_` to match the existing keys we had before the design doc.

Example of a valid config structure:

```toml
active_context = "my-context" # Currently active context (optional)

# Any optional option is managed under preferences.
[preferences]
location = "fsn1" # Global default location preference

[preferences.sort]
certificate = "id:asc"

[[contexts]]
name = "my-context"
token = "def"

[contexts.preferences] # Confusingly, this belongs to the "my-context" context in TOML
location = "hel1" # Override for the global "location" option

# Examples of namespaced preferences:
[contexts.preferences.sort]
server = "created:asc"
load_balancer = "id:desc"

[contexts.preferences.columns]
server = "name,id,age"
load_balancer = "$defaults,age,!id" # Some options to expand the columns syntax in the future
```

### How do we persist the configuration file?

<aside>
🗣️ This section only describes the current state. We do not plan an making any changes, besides adding a new `--config` flag.

</aside>

The configuration file is encoded as `TOML`. By default we look in these paths for the file:

- Default config path:
    - Unix: `~/.config/hcloud/cli.toml`
    - Windows: `%APPDATA%/hcloud/cli.toml`

The config path can be overridden using the `HCLOUD_CONFIG` env variable or the `--config` global flag.

When the user makes any changes, we re-write the full file. We do not take any precautions to persist user formatting or comments. This makes it way easier for us to write out any changes back into the file.

### Environment Variables

We want to allow users to set all options through environment variables.

Our environment variables always start with `HCLOUD_` and then use the option key in uppercase and with `-` and `.` replaced with `_`.

- Examples:
    - `token` → `HCLOUD_TOKEN`
    - `default-ssh-keys` → `HCLOUD_DEFAULT_SSH_KEYS`
    - `sort.server` → `HCLOUD_SORT_SERVER`

### Flags

We want to allow users to set options through flags.

The flag names are the option key with `.` replaced with `-`.

- Examples:
    - `token` → `--token` (only example, flag not present in CLI)
    - `quiet` → `--quiet`
    - `namespace.foo` → `--namespace-foo`

We will not add global flags for options that already have flags in sub-commands (See above).

### Precedence

The value for an option can be set in many different places. We apply the following precedence to decide which value will be used.

1. Interactive Mode (not yet implemented)
2. Flags
3. Environment Variables
4. Configuration File
    1. Context-specific
    2. Global
5. CLI Defaults

The order matches how “far” the user is from where the option was set, with “further” sources having a lower priority.

⇒ The user is very near and interacts directly with the interactive mode or flags. Environment variables have been set before running the command. The configuration file is in another directory altogether and mostly hidden from the user. CLI Defaults are statically set by us in code, the user has no direct influence on this.

### Backwards Compatible Options

Some options currently do not have consistent flags/config file keys/environment variables. We do not want to break the existing usages. This section documents any options that deviate from the ideal naming scheme.

- Option Key: `token`
    - Env: `HCLOUD_TOKEN` (*Already exists*)
    - Flag: Not present because of security concerns ⚠️
    - Config: `token` (*Already exists*)
- Option Key: `context`
    - Env: `HCLOUD_CONTEXT` (*Already exists*)
    - Flag: `--context`
    - Config: `active_context` (*Already exists*) ⚠️
- Option Key: `config`
    - Env: `HCLOUD_CONFIG` (*Already exists*)
    - Flag: `--config`
    - Config: We do not plan to add this, as it is redundant and unclear what it would mean. ⚠️

### Documentation

We plan to add a new sub-command `hcloud help config`. In there we will:

- Explain what options are
- Explain exceptions to schema (`--token` flag not present, `HCLOUD_CONTEXT` vs `active_context` etc.)
- Explain where users can set the options
- Explain the renaming of configs between env variables, flags and the config file
- List all available options with a short explanation on their purpose.

# Release Strategy

We will cut a normal CLI Release once this is merged and available for users.

# Alternatives

### Alternative configuration commands: User edits config file themselves

→ **Benefits:** Less work for us

→ **Why we decided against it:** Bad UX, no validation / feedback for the user.

### Alternative configuration commands: Specialized sub-commands for every option

- Example: `hcloud context default-ssh-keys add <ssh-key>`

→ **Benefits:** Intuitive command layout, easy customization

→ **Why we decided against it:** Lots of clutter, would need a lot of documentation, more effort to maintain

### Alternative handling of flags: All flags are handled by viper

- Every flag becomes an option
    - Example: `hcloud server create --location` flag becomes `server.create.location` option
- Options can be defined in the config, via environment variables or via flags
- Because of the hierarchy, usage stays consistent, with the difference that you can now define defaults for flags using the configuration file or environment variables

→ **Benefits:** Only one source of truth (Viper) and every flag is configurable

→ **Why we decided against it:** Requires major rewrite, very complex implementation for little benefit, no shared options between commands

### **Alternative config file schema: Namespaced option keys saved flat**

→ **Benefits:**

- Easier implementation

→ **Why we decided against it**

- Nested objects provide better structure for configuration file

### Alternative config command name: `hcloud context config`

→ **Benefits:** Makes apparent that configuration happens on a context level

→ **Why we decided against it:** Bad UX

phm07:
This PR implements the new configuration system as described in #762.

Closes #762
In preparation for #434


internal/cmd/firewall/describe.go
```diff
@@ -59,7 +59,7 @@ var DescribeCmd = base.DescribeCmd{
 				var ips []net.IPNet
 				switch rule.Direction {
 				case hcloud.FirewallRuleDirectionIn:
-					cmd.Print("    Source IPs:\n")
+					cmd.Print("    Origin IPs:\n")
```
apricote, line 62: Is this change related to the PR or did it accidentally slip in?

internal/state/config/config.go
```diff
@@ -1,178 +1,248 @@
...
-		return err
+	cfg.contexts = cfg.schema.Contexts
+	for i, ctx := range cfg.contexts {
+		if ctx.ContextName == activeContext {
+			cfg.activeContext = cfg.contexts[i]
+			break
+		}
 	}
-	if err := os.MkdirAll(filepath.Dir(cfg.path), 0777); err != nil {
-		return err
```
jooola, line 67: Happy not to see this 0777 in the new code, but shouldn't we keep the config dir creation behavior?

internal/state/config/options.go
```diff
@@ -0,0 +1,278 @@
...
+	"strings"
+	"time"
+
+	"github.com/spf13/pflag"
+
+	"github.com/hetznercloud/cli/internal/cmd/util"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+)
+
+type OptionFlag int
```
jooola, line 16: Should we rename this to `OptionFeature`? I have and will be confused by this naming more than once.

e.g. `!o.HasFlag(OptionFlagEnv)` => `!o.HasFeature(OptionFeatureEnv)`

jooola (commented):
Overall, I like how it is structured.

I think a few refactoring passes are needed before this can be merged.

I also noticed that the viper library already ships our previous toml library, so might be shipping both. Are we ok with the binary size?

apricote (commented):
Partial review (its a huge PR ;) ), will continue tomorrow :)

internal/state/config/preferences.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/pelletier/go-toml/v2"
+	"github.com/spf13/viper"
+)
+
+// Preferences are options that can be set in the config file, globally or per context
+type Preferences map[string]any
+
+func (p Preferences) Get(key string) (any, bool) {
+	configKey := strings.ReplaceAll(strings.ToLower(key), "-", "_")
```
apricote, line 16: The `Get`, `Set` and `Unset` methods share a lot of common code related to the config code and path splitting. Is there some way to make them share the behaviour?

internal/state/config/preferences.go
```diff
@@ -0,0 +1,116 @@
...
+	var m map[string]any = p
+	path := strings.Split(configKey, ".")
+	for i, key := range path {
+		if i == len(path)-1 {
+			val, ok := m[key]
+			return val, ok
+		}
+		if next, ok := m[key].(map[string]any); !ok {
+			break
```
apricote, line 25: Should this be reported as a warning/corrupt config file?

internal/state/config/preferences_test.go
```diff
@@ -0,0 +1,171 @@
...
+	assert.True(t, p.Unset("quux.garply.xyzzy"))
+	assert.Equal(t, config.Preferences{
+		"baz": "qux",
+	}, p)
+
+	assert.True(t, p.Unset("baz"))
+	assert.Equal(t, config.Preferences{}, p)
+}
+
+func TestPreferences_Validate(t *testing.T) {
```
apricote, line 150: Can you add a test for nested options?

internal/state/config/preferences_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/hetznercloud/cli/internal/state/config"
+)
+
+func TestPreferences_Get(t *testing.T) {
```
apricote, line 11: Can you add a test for a missing nested option?

internal/cmd/config/add.go
```diff
@@ -0,0 +1,101 @@
...
+		ValidArgsFunction: cmpl.NoFileCompletion(cmpl.SuggestArgs(
+			cmpl.SuggestCandidatesF(func() []string {
+				var keys []string
+				for key, opt := range config.Options {
+					if opt.HasFlag(config.OptionFlagPreference) {
+						keys = append(keys, key)
+					}
+				}
+				return keys
+			}),
```
apricote, line 35: Is there any way to only suggest list options?

internal/cmd/config/add.go
```diff
@@ -0,0 +1,101 @@
...
+
+	prefs.Set(key, val)
+
+	if len(added) == 0 {
+		_, _ = fmt.Fprintln(os.Stderr, "Warning: no new values were added")
+	} else if len(added) < len(values) {
+		_, _ = fmt.Fprintln(os.Stderr, "Warning: some values were already present or duplicate")
+	}
+
+	if util.IsNil(ctx) {
```
apricote, line 95: You could also branch on `global` here.

internal/cmd/config/get.go
```diff
@@ -0,0 +1,60 @@
...
+	for name, o := range config.Options {
+		if name == key {
+			opt = o
+			break
+		}
+	}
+	if opt == nil {
+		return fmt.Errorf("unknown key: %s", key)
+	}
+
```
apricote, line 53:
```suggestion
	key := args[0]
	
	opt := config.Options[key]
	if opt == nil {
		return fmt.Errorf("unknown key: %s", key)
	}

```

internal/cmd/config/get.go
```diff
@@ -0,0 +1,60 @@
...
+	global, _ := cmd.Flags().GetBool("global")
+	allowSensitive, _ := cmd.Flags().GetBool("allow-sensitive")
+
+	if global {
+		// set context to nil and then reload
+		config.OptionContext.OverrideAny(s.Config(), nil)
+		s.Config().Reset()
+		if err := config.ReadConfig(s.Config(), nil); err != nil {
+			return err
+		}
```
apricote, line 39: This code is duplicated between `get.go` and `list.go`. I think this makes more sense to have in `state/config/config.go`. That way `config.Reset()` could also be dropped from the interface.

internal/cmd/config/set.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/hetznercloud/cli/internal/cmd/util"
+	"github.com/hetznercloud/cli/internal/state"
+	"github.com/hetznercloud/cli/internal/state/config"
+)
+
+func NewSetCommand(s state.State) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "set <key> <value>...",
+		Short:                 "Set a configuration value",
+		Long:                  "Set a configuration value. For a list of all available configuration options, run 'hcloud help config'.",
```
apricote, line 21: I think we should document the accepted values for durations. This is not obvious for non-go devs.

apricote (changes requested):

internal/cmd/config/set.go
apricote: Adding a quick description in either `hcloud config --help` or `hcloud config set --help` sounds good. While this is very intuitive, you need to know that it exists in the first place. People could expect that this is seconds, milliseconds, ...

internal/state/config/config.go
```diff
@@ -1,178 +1,294 @@
...
-func (cfg *config) ActiveContext() *Context {
-	return cfg.activeContext
-}
+func (cfg *config) Write(w io.Writer) (err error) {
+	if w == nil {
+		dir := path.Dir(cfg.path)
+		if err = os.MkdirAll(dir, 0750); err != nil {
+			return err
+		}
+		f, err := os.OpenFile(cfg.path, os.O_WRONLY|os.O_CREATE|os.O_APPEND|os.O_TRUNC, 0600)
```
jooola, line 199: I am not sure if the append flag is needed here.

jooola (commented):
Took another look.

Do you think it would be possible to add the `config.OptionQuiet` options inside the config struct? This way we would only have the config struct in the state and get the value from it?

So instead of `config.OptionQuiet.Get(s.Config())` we would simply have `s.Config().Quiet()` ?


Other than that, there is a bunch of code that can be deduplicated, extracted in smaller functions, so the higher level logic is easier to understand.

internal/state/config/options.go
jooola: Alright, maybe with some getting used to I'll get the hand of it.

internal/cmd/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+			`
+Since the above options are not preferences, they cannot be modified with 'hcloud config set' or 
+'hcloud config unset'. However, you are able to retrieve them using 'hcloud config get' and 'hcloud config list'.
+Following options are preferences and can be used with these commands:
+
+` + preferenceOptions +
+			`
+Options will be persisted in the configuration file. To find out where your configuration file is located
+on disk, run 'hcloud config get config'.
+`,
```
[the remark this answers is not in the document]
jooola: What I had in mind was to maybe embed an entire markdown file, that will be easier to maintain in the long run, and render it in the terminal (e.g. https://github.com/charmbracelet/glamour).

Maybe using some template function, we can embed the preferences as Markdown table as well.

jooola:
> There would be a lot of code duplication. There would need to be a method implemented on the config struct for every single option. Also there would need to be a map that stores all option. There is no way to create such a map automatically when getter methods are used to access options. Right now newOpt is used to handle that.

I don't think it increases the duplication too much, it is moving some code around. Instead of having "complex" code in the business logic, we have dumb code in the config package.

> Testing would be a lot harder since options would be hard coded. Currently, the option set is extended for some tests. This is for example used to test nested options which don't exist yet.

I am not sure if this is a real problem, we could have dedicated config struct for testing that embed the normal config struct. 

> It would unnecessarily clutter the Config interface.

I prefer to clutter the config interface instead of the business code. Ensuring a proper separation of concern with simple APIs, will make the project easier to work with.

> Feel free to comment those parts in your review.

Will do.

jooola:
> There would be a lot of code duplication. There would need to be a method implemented on the config struct for every single option. Also there would need to be a map that stores all option. There is no way to create such a map automatically when getter methods are used to access options. Right now newOpt is used to handle that.

That said, I don't have a strong opinion. I just wanted to discuss the topic, to see if we can have an even better API.

internal/cmd/config/add.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if global {
+		prefs = s.Config().Preferences()
+	} else {
+		ctx = s.Config().ActiveContext()
+		if util.IsNil(ctx) {
+			return fmt.Errorf("no active context (use --global flag to set a global option)")
+		}
+		prefs = ctx.Preferences()
+	}
```
jooola, line 66: I saw this appear multiple times in the code, might be a candidate for a dedicated function.

internal/cmd/config/get.go
```diff
@@ -0,0 +1,51 @@
...
+		Args:                  util.Validate,
+		TraverseChildren:      true,
+		DisableFlagsInUseLine: true,
+		SilenceUsage:          true,
+		RunE:                  state.Wrap(s, runGet),
+	}
+	cmd.Flags().Bool("global", false, "Get the value globally")
+	cmd.Flags().Bool("allow-sensitive", false, "Allow showing sensitive values")
+	return cmd
+}
```
jooola, line 27: We do not provide config key completion here?

internal/cmd/config/add.go
```diff
@@ -0,0 +1,101 @@
...
+
+	key, values := args[0], args[1:]
+	opt, ok := config.Options[key]
+	if !ok || !opt.HasFlag(config.OptionFlagPreference) {
+		return fmt.Errorf("unknown preference: %s", key)
+	}
```
jooola, line 72: I saw this appear multiple times in the code, might be a candidate for a dedicated function.

internal/cmd/config/remove.go
```diff
@@ -0,0 +1,104 @@
...
+				}
+				return keys
+			}),
+			cmpl.SuggestCandidatesCtx(func(_ *cobra.Command, args []string) []string {
+				var comps []string
+				if opt, ok := config.Options[args[0]]; ok {
+					comps = opt.Completions()
+				}
+				return comps
+			}),
```
jooola, line 42: I saw this appear multiple times in the code, might be a candidate for a dedicated function.

internal/cmd/config/remove.go
```diff
@@ -0,0 +1,104 @@
...
+
+	switch opt.T().(type) {
+	case []string:
+		before := util.AnyToStringSlice(val)
+		diff := util.SliceDiff[[]string](before, values)
+		val = diff
+		removed = util.ToAnySlice(util.SliceDiff[[]string](before, diff))
+	default:
+		return fmt.Errorf("%s is not a list", key)
+	}
```
jooola, line 84: I saw this appear multiple times in the code, might be a candidate for a dedicated function.

internal/cmd/config/set.go
```diff
@@ -0,0 +1,121 @@
...
+			return fmt.Errorf("expected exactly one value")
+		}
+		value := values[0]
+		switch strings.ToLower(value) {
+		case "true", "t", "yes", "y", "1":
+			val = true
+		case "false", "f", "no", "n", "0":
+			val = false
+		default:
+			return fmt.Errorf("invalid boolean value: %s", value)
```
jooola, line 87: strconv.ParseBool ?

internal/cmd/config/list.go
```diff
@@ -0,0 +1,92 @@
...
+		Key   string `json:"key"`
+		Value any    `json:"value"`
+	}
+
+	var options []option
+	for name, opt := range config.Options {
+		val := opt.GetAsAny(s.Config())
+		if opt.HasFlag(config.OptionFlagSensitive) && !allowSensitive {
+			val = "[redacted]"
+		}
```
jooola, line 57: Should the 'redacted' logic be moved to the option methods ? And have a dedicated method when allowSensistive is set?

internal/cmd/config/set.go
```diff
@@ -0,0 +1,121 @@
...
+	case time.Duration:
+		if len(values) != 1 {
+			return fmt.Errorf("expected exactly one value")
+		}
+		value := values[0]
+		var err error
+		val, err = time.ParseDuration(value)
+		if err != nil {
+			return fmt.Errorf("invalid duration value: %s", value)
+		}
```
jooola, line 103: Does it make sense to move the Parsing/Formatting logic in the Option?

jooola (commented):
This is getting better and better, I like it!

I added the few comment in places I though we could deduplicate/move some code.

internal/cmd/config/set.go
jooola: Then I believe this could also be a candidate for a dedicated function :)

jooola: To improve readability and split concerns: if I read code about the set command, I don't to see 6 lines only dedicated to parsing a bool value. Having a single function can make the code way more readable.

jooola:
> `config.OptionQuiet.Get(s.Config())` => `s.Config().Quiet()`

Alright, case close from my side then.

README.md
```diff
@@ -142,18 +142,31 @@ You can control output via the `-o` option:
...
-* `HCLOUD_CONTEXT`
-* `HCLOUD_CONFIG`
-
-When using `hcloud` in scripts, for example, it may be cumbersome to work with
-contexts. Instead of creating a context, you can set the token via the `HCLOUD_TOKEN`
-environment variable. When combined with tools like [direnv](https://direnv.net), you
-can configure a per-directory context by setting `HCLOUD_CONTEXT=my-context` via `.envrc`.
+## Configuring hcloud
+
+The hcloud CLI tool allows configuration using following methods:
```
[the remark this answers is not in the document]
jooola: Yes work for me.

apricote (approved):
Awesome! Looking really good now :)

(We will _for sure_ find some bugs and rough edges after merging, but that happens always).

PR #740, opened 2024-04-25
chore: clean up tooling

phm07:
Instead of installing tools globally, they should be imported in tools.go and then used with `go run`. This has the following advantages:
- Doesn't clutter your dev environment
- Ensures correct versions of tools are used
- Tooling can be updated by renovate


apricote (approved):

jooola (changes requested):
One GitLab CI job hasn't been cleaned.

jooola (approved):
Only a `go mod tidy` is missing and it's good to go!

PR #759, opened 2024-05-21
feat(server): add default-ssh-keys option

phm07:
This PR adds the `default-ssh-keys` option which allows you to define a list of SSH keys to be used as defaults for the `hcloud server create` command.


apricote (approved):
Awesome!

PR #773, opened 2024-06-05
ci: improve goreleaser pipeline

phm07:
This PR improves the goreleaser pipeline in the following ways:
- It utilizes the [YAML merge key](https://yaml.org/type/merge.html) to reduce code duplication
- It adds the `-s` ldflag to reduce binary size: [Omit the symbol table and debug information.](https://pkg.go.dev/cmd/link)
- It adds the `-trimpath` flag to trim absolute file paths from build output

Binary size of `linux/amd64` build on:
- 069fffe48bb09f23edcade9e73f351a8b361a2a5: 14192497 B
- d0dfda9dd855c25a6815019f8dc72c07dff69fb6: 12738712 B

-> File size reduction is about 10%.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -3,42 +3,28 @@ before:
     - go mod tidy
 
 builds:
-  - id: hcloud-build
-    main: ./cmd/hcloud/main.go
-    binary: hcloud
-    ldflags:
-      - -w
-      - -X {{ .ModulePath }}/internal/version.version={{ .Version }}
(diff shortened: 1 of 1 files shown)


jooola (approved):
Awesome!

Small note on the YAML merge keys, the feature is only part of YAML 1.1 and is born deprecated. YAML 1.0 and 1.2 does not have merge keys.

jooola:
Does stripping the symbol table influence traces from a panic?

apricote:
I am personally against stripping any debug information from binaries, as it makes later analysis/profiling harder. As we already strip the DWARF section (`-w`), I guess that would be another discussion altogether and I am fine with adding the `-s` here for now.

PR #785, opened 2024-06-19
test(e2e): add end to end testing

phm07:
This PR adds end to end tests. Tests are located in `test/e2e/`. The `HCLOUD_TOKEN` environment variable should contain an API token to an empty project for the tests to work.

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -28,9 +28,8 @@ func TestNetwork(t *testing.T) {
 
 	_, err = createNetwork(t, networkName, "--ip-range", "10.0.1.0/24")
 	require.Error(t, err)
-	// The API currently (wrongly) returns service_error instead of uniqueness_error.
-	// Once this is fixed we have to change it here.
-	assert.Regexp(t, `^name is already used \(service_error, [0-9a-f]+\)$`, err.Error())
+	// TODO: API currently returns service_error instead of uniqueness_error. Add this back in once this is fixed.
+	// assert.Regexp(t, `^name is already used \(uniqueness_error, [0-9a-f]+\)$`, err.Error())
 
 	t.Run("enable-protection", func(t *testing.T) {
 		t.Run("non-existing-protection", func(t *testing.T) {


e2e_test/combined_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCombined(t *testing.T) {
+	// combined tests combine multiple resources and can thus not be run in parallel
+	serverID := createServer(t, "test-server", "cpx11", "ubuntu-24.04")
```
apricote, line 14: Can we move constants like the server type and image to variables? This makes them easier to replace if the resource is deprecated/removed in the future or we want to use something else for better utilization

Similar to how its done in Terraform: https://github.com/hetznercloud/terraform-provider-hcloud/blob/ac90449bdd4782e0c58930b638cb3e8f4e313725/internal/teste2e/testing.go#L19-L43

e2e_test/combined_test.go
```diff
@@ -0,0 +1,82 @@
...
+	out, err = runCommand(t, "floating-ip", "describe", strconv.Itoa(floatingIP))
+	require.NoError(t, err)
+	assert.Regexp(t, `ID:\s+[0-9]+
+Type:\s+ipv4
+Name:\s+test-floating-ip
+Description:\s+-
+Created:.*?
+IP:\s+(?:[0-9]{1,3}\.){3}[0-9]{1,3}
+Blocked:\s+no
+Home Location:\s+[a-z]{3}[0-9]+
```
apricote, line 59: Not all locations have the number suffix. Specifially `ash`, `hil` and `sin` are missing it.

.github/workflows/ci.yml
```diff
@@ -33,6 +33,8 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
apricote, line 37: The permission is no longer necessary for this job (was required for TPS).

.github/workflows/e2e_test.yml
```diff
@@ -0,0 +1,33 @@
+name: E2E Tests
```
apricote, line 1: This workflow is missing the `GO_TOOLCHAIN=local` env.

In general I consider e2e tests to be a part of the CI, so I would move them to `ci.yml`.

.github/workflows/e2e_test.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Run tests
+        run: go test -tags e2e -coverpkg=./... -coverprofile=coverage.txt -v -race ./e2e_test
+
+      - name: Upload coverage reports to Codecov
+        if: >
+          !startsWith(github.head_ref, 'renovate/') &&
+          !startsWith(github.head_ref, 'release-please--')
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
apricote, line 33: Needs flags so codecov can properly show that these are two different kinds of coverage. Please also add a `flags: unit` to the unit test coverage upload.

```suggestion
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          flags: e2e
```

e2e_test/combined_test.go
```diff
@@ -0,0 +1,82 @@
+package e2e_test
```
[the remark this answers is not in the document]
apricote: I dont think moving it to `internal/` is necessary per se. All of this is `_test.go` files, so they cant be imported anyway.

`test/e2e` is for consistency with our other repos (csi-driver, hccm)

e2e_test/network_test.go
```diff
@@ -0,0 +1,331 @@
...
+	assert.Empty(t, out)
+	require.EqualError(t, err, `required flag(s) "ip-range", "name" not set`)
+
+	networkName := withSuffix("test-network")
+	networkID, err := createNetwork(t, networkName, "--ip-range", "10.0.0.0/24")
+	require.NoError(t, err)
+
+	_, err = createNetwork(t, networkName, "--ip-range", "10.0.1.0/24")
+	require.Error(t, err)
+	assert.Regexp(t, `^name is already used \(uniqueness_error, [0-9a-f]+\)$`, err.Error())
```
apricote, line 31: Can we disable this assertion until the API response is fixed? To validate that the CI & Coverage works as expected

apricote: Can you comment this out instead of checking for the wrong response? Otherwise the tests start failing once this is fixed.

apricote (approved):

PR #793, opened 2024-06-26
fix: debug log is truncated if it already exists

### LINKED ISSUE #763: bug: debug log is truncated if it already exists
If `HCLOUD_DEBUG` is enabled and `HCLOUD_DEBUG_FILE` is set, the log file is truncated if it already exists. This leads to previous logs being lost. Instead, it should append to the file. (Maybe also log the command that was run?)

phm07:
This fixes #763. It also writes the used command to the debug logs for easier debugging.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -3,6 +3,7 @@ package state
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/state/state.go
```diff
@@ -66,8 +67,9 @@ func (c *state) newClient() hcapi2.Client {
 		if filePath := config.OptionDebugFile.Get(c.config); filePath == "" {
 			opts = append(opts, hcloud.WithDebugWriter(os.Stderr))
 		} else {
-			writer, _ := os.Create(filePath)
-			opts = append(opts, hcloud.WithDebugWriter(writer))
+			f, _ := os.OpenFile(filePath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```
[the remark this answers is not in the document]
apricote: Could you add an error handler here? We could either just log it to stderr an continue, or abort the call

apricote (approved):

internal/state/state.go
```diff
@@ -66,8 +67,9 @@ func (c *state) newClient() hcapi2.Client {
...
+			_, _ = f.WriteString("--- Command:\n" + strings.Join(os.Args, " ") + "\n\n\n\n")
```
jooola, line 71: Since we are adding this, do you think it's worth quoting the args to detect quoting issues?

internal/state/state.go
```diff
@@ -88,8 +90,21 @@ func (c *state) newClient() (hcapi2.Client, error) {
...
+			}
+
+			quotedArgs := make([]string, 0, len(os.Args))
+			for _, arg := range os.Args {
+				quotedArgs = append(quotedArgs, fmt.Sprintf("%q", arg))
+			}
+			_, err = f.WriteString("--- Command:\n" + strings.Join(quotedArgs, " ") + "\n\n\n\n")
+			if err != nil {
+				return nil, err
+			}
```
jooola, line 105: This only happens for `HCLOUD_DEBUG_FILE`, not for `HCLOUD_DEBUG`. We might need to break up the condition.

jooola (approved):

Issue #418, opened 2022-11-13
Feature request: allow sorting to be configured globally


cedi:
I am a heavy user of the hcloud-cli. However, I really don't like the default sorting by server-id, and would prefer sorting by age (see #417 where I request adding a computed "age" column) or name.

Currently, I always type `hcloud server list -s name` to sort by name.

since the `-s` flag is part of the `list` subcommand and not the `hcloud` base command, I cannot simply `alias hcloud-"hcloud -s name"` in my shell to default the sorting to "by name".
I think this could be implemented as part of the configuration toml.
I would expect the following behavior:

|config file|flag|result|
|-|-|-
|not-set|not-set|sort by id (default)|
|not-set|set|sort by flag value|
|set|not-set|sort by config value|
|set|set|sort by flag value|

I would also volunteer to implement this myself and submit a PR, but I wanted to gauge interest first, before putting in some work and having the PR rejected.

cedi:
@apricote bump

PR #795, opened 2024-06-26
feat: allow configuring default sort columns for each supported resource

phm07:
This PR adds options to configure the default sorting column for each resource that supports the `list` command.
See #418 and #434

### DIFF
--- a/internal/cmd/base/list.go
+++ b/internal/cmd/base/list.go
@@ -2,6 +2,7 @@ package base
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -64,8 +65,14 @@ func (lc *ListCmd) Run(s state.State, cmd *cobra.Command) error {
 		PerPage:       50,
 	}
 
-	sorts, _ := cmd.Flags().GetStringSlice("sort")
-	if lc.SortOption != nil && !cmd.Flags().Changed("sort") {
+	var sorts []string
+	if cmd.Flags().Changed("sort") {
+		if lc.SortOption == nil {
+			_, _ = fmt.Fprintln(os.Stderr, "Warning: resource does not support sorting. Ignoring --sort flag.")
+		} else {
+			sorts, _ = cmd.Flags().GetStringSlice("sort")
+		}
+	} else if lc.SortOption != nil {
 		var err error
 		sorts, err = lc.SortOption.Get(s.Config())
 		if err != nil {


jooola:
Do you think it's possible to decouple the `SortOption` and the per resource sort arguments? Similar to the tables columns?

jooola:
> Is this too much? Or should we keep all the entries fort sort.[resource]?

I'd rather keep it explicit, unless we have a clear list of resource names.

> sort.loadbalancertype │ HCLOUD_SORT.LOADBALANCERTYPE

Could we keep the `_` or `-` in the names? The env variable name is not the final ones right?

jooola:
Nope, env vars cannot have dots: 

Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2017 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names. Uppercase and lowercase letters shall retain their unique identities and shall not be folded together. The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.


https://pubs.opengroup.org/onlinepubs/9699919799/

apricote:
Without taking a look at the code:

- I think the options should be the same as the subcommands: `hcloud load-balancer` -> `sort.load-balancer`
- In the docs I would prefer to have a single element `sort.<resource>` to keep this reasonably small.
- I would add a section in `hcloud config --help` that describes these options in general

apricote (approved):
:rocket:

apricote:
While testing this, I found that `hcloud config add` adds to the front of the list. I found this counterintuitive. I wanted to set a default for `hcloud image list --sort name:asc --sort created:asc` so I ran:

```
$ hcloud config add sort.image name:asc
$ hcloud config add sort.image created:asc
```

Thinking it would end us as `[name:asc created:asc`]. Turned out it was the other way around, and running `hcloud image list` now orders the other way around.

```
$ hcloud config get sort.image
[created:asc name:asc]
```


---

So feature works technically, but the UX was unexpected

apricote:
Even with `hcloud config set sort.image name:asc created:asc` its being saved in the wrong order.

internal/state/config/options.go
```diff
@@ -142,6 +146,134 @@ var (
...
+		nil,
+	)
+
+	OptionSortISO = newOpt(
+		"sort.iso",
+		"Default sorting for ISO resource",
+		[]string{"id:asc"},
+		(DefaultPreferenceFlags&^OptionFlagPFlag)|OptionFlagSlice|OptionFlagHidden,
+		nil,
+	)
```
apricote, line 196: I noticed that the API does not support sorting for:

- ISOs
- Load Balancer Types
- Networks
- Server Types

I think we should remove these options and only add them if the API also supports this.

apricote (changes requested):
I noticed that this PR does not compile anymore because of #798. Could you rebase this PR and fix the compilation error?

The functionality works great now!

apricote (approved):
:rocket:

PR #798, opened 2024-06-27
feat: better error messages on malformed option value

phm07:
Previously, if an option was configured with the wrong type (e.g. `string` was configured with an `int` value) or other parsing errors like malformed duration/boolean values occured, `Option[T].Get()` would panic. It now returns an error that can properly be displayed to the user and falls back to the default value.

### DIFF
--- a/internal/state/config/options.go
+++ b/internal/state/config/options.go
@@ -157,7 +157,7 @@ func (o *Option[T]) Get(c Config) (T, error) {
 	// Since viper uses multiple configuration sources (env, config, etc.) we need
 	// to be able to convert the value to the desired type.
 	val := c.Viper().Get(o.Name)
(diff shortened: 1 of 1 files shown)


apricote (commented):
Is there any way we can validate the config once at startup instead of doing this whenever one wants to access an option?

apricote:
> We could run this method one time for each option on startup (or on the first Get() call if the cache is not yet initialized) and then cache the result. Then we would only need a static type cast after that for accessing the value.
However, it would make the code a lot more complex and the performance difference would not be noticeable (probably not even measurable), so I don't know if it's worth it.

My primary concern is not the performance, but rather that it is very annoying to check for an error every time you access the config.

apricote (approved):
As much as I would like to see the error handling to be gone, I think we can merge this for now.

PR #800, opened 2024-06-28
fix: correctly print disk sizes on 32bit systems

jooola:
The order of when the casting occurs is important.

Can be the bug reproduced with:
```
GOARCH=386 go test  ./...
```


jooola:
Found while updating the hcloud apline package: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/68362

jooola:
```go
package main

import (
	"fmt"

	"github.com/dustin/go-humanize"
)

func main() {
	size := int(10)

	fail := uint64(size * humanize.GByte)
	happy := uint64(size) * humanize.GByte

	fmt.Printf("fail: %v\n", fail)
	fmt.Printf("fail: %s\n", humanize.Bytes(fail))

	fmt.Printf("happy: %v\n", happy)
	fmt.Printf("happy: %v\n", humanize.Bytes(happy))
}
```


```sh
$ go run ./main.go
fail: 10000000000
fail: 10 GB
happy: 10000000000
happy: 10 GB

$ GOARCH=386 go run ./main.go
fail: 1410065408
fail: 1.4 GB
happy: 10000000000
happy: 10 GB

```

apricote (approved):

PR #812, opened 2024-07-10
fix(firewall): wrong wording when firewall is removed from resource

### LINKED ISSUE #809: Working when a firewall is removed from resource is wrong
### TL;DR

The wording when removing a firewall from a resource is wrong.

### Expected behavior

The wording should match the result of the action that was taken.

### Observed behavior

https://github.com/hetznercloud/cli/blob/1ac27bfbcca1eb3dbb3778b1894badff3d7462f2/internal/cmd/firewall/remove_from_resource.go#L86

### Minimal working example

_No response_

### Log output

_No response_

### Additional information

_No response_

phm07:
Closes #809

### DIFF
--- a/internal/cmd/firewall/apply_to_resource_test.go
+++ b/internal/cmd/firewall/apply_to_resource_test.go
@@ -43,7 +43,7 @@ func TestApplyToServer(t *testing.T) {
 
 	out, errOut, err := fx.Run(cmd, []string{"--type", "server", "--server", "my-server", "test"})
 
-	expOut := "Firewall 123 applied\n"
+	expOut := "Firewall 123 applied to resource\n"
 
 	assert.NoError(t, err)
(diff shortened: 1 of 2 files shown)


jooola (approved):

jooola (approved):

internal/cmd/firewall/apply_to_resource_test.go
```diff
@@ -43,7 +43,7 @@ func TestApplyToServer(t *testing.T) {
 
 	out, errOut, err := fx.Run(cmd, []string{"--type", "server", "--server", "my-server", "test"})
 
-	expOut := "Firewall 123 applied\n"
+	expOut := "Firewall 123 applied to resource\n"
```
jooola, line 46: We might target many resources when using the label selector, but I think this is good enough.

PR #815, opened 2024-07-11
ci: add coverpkg flag for testing

phm07:
This PR adds the `-coverpkg=./...` flag so that coverage is not only restricted to lines in the same package as the running test. This flag is needed for end to end testing. Adding the flag now makes it easier for us to compare the coverage of e2e tests with the main branch.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -47,7 +47,7 @@ test:unit:
   stage: test
   image: golang:1.21
   script:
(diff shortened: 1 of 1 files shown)


apricote (approved):

jooola (changes requested):
We are missing an update for the Gitlab CI config.

jooola (approved):

PR #839, opened 2024-08-12
feat: allow deleting rDNS entries

### LINKED ISSUE #814: fix: allow deleting reverse DNS entries
At the moment, `set-rdns` only allow adding reverse DNS entries, but they don't allow deleting them. This should be made possible because the API supports it.

phm07:
This PR adds a `--reset` flag to the `set-rdns` commands which allows passing a null value as the hostname. This resets the reverse DNS entry to the original entry for the IP.

Fixes #814

### DIFF
--- a/internal/cmd/base/set_rdns.go
+++ b/internal/cmd/base/set_rdns.go
@@ -19,7 +19,6 @@ type SetRdnsCmd struct {
 	ResourceNameSingular string // e.g. "server"
 	ShortDescription     string
 	NameSuggestions      func(client hcapi2.Client) func() []string
-	AdditionalFlags      func(*cobra.Command)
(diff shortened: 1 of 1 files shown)


internal/cmd/base/set_rdns.go
```diff
@@ -39,7 +39,7 @@ func (rc *SetRdnsCmd) CobraCommand(s state.State) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringP("hostname", "r", "", "Hostname to set as a reverse DNS PTR entry (required)")
```
apricote, line 41: The docs for this still include "(required)".

Does this need an update to the `Use` line to have both alternatives specified?

apricote (changes requested):

apricote (approved):

Issue #613, opened 2023-11-15
Option to disable wait for apply_firewall


simonostendorf:
I think it would be nice to have a command flag that disables the wait for `apply_firewall`.

I have a script that uses the hcloud cli to adjust my firewall rules if a new node joins my cluster. The script should not wait that the new rules are applied to each cluster node because the label selector takes care of that.

pauhull:
Would it not be possible to run the command in the background (using `&`)?

simonostendorf:
> Would it not be possible to run the command in the background (using `&`)?

Works for now. Closing the issue.

Issue #284, opened 2020-11-04
run command async without foreground


omerh:
Hi,

Just trying out the cli tool, I was wondering if there is an option to run it without `--poll-interval` ?
I want to run a snapshot script and I don't need to see the progress of the snapshot creation.

I tough of `--poll-interval -1` or something

LKaemmerling:
Currently, there is no way to disable the progress bar at the moment. 

Do i understand it correctly, that you just want to disable the progress bar, but the command should still wait until the snapshot is created?

Specifying the poll-interval should only be used when you run into the rate limit.

omerh:
Hi,
Not exactly, I don't want to see the progress and don't want to wait until its finished.
Just send an async request for snapshot 

I've made this https://github.com/omerh/hetzner-hcloud-snapshots and I don't need to wait till its done

LKaemmerling:
But then you will never know if the snapshot is ready (useable) or not?

omerh:
Im counting on Hetzner to succeed in snapshot (like I do in AWS and DO) 
If it may break than you can leave it and close the issue 👍

ktomk:
For me with hcloud cli it works pretty well redirecting stdout to /dev/null - this effectively removes the progress-bar.

Stderr is untouched as I normally want to see the errors (could be misconfiguration / changes as well), when unattended all output is also turned into a notification, so I only want the errors.

Can imagine that to prevent the output of the progress bar and run in background in your script could do:
~~~
hcloud server create-image --type snapshot --label id=$1 \
     --description server-$1 --poll-interval 0 $1 >/dev/null &
                                                  ########## #
~~~

Is this what you're looking for?

omerh:
Yes. But buitin the command and not directing to `/dev/null`

LKaemmerling:
I would still recommend directing it to `/dev/null`. Your "use case" is really limited, and you should always expect that there might be a problem or something else and then react to it.  Therefore i think given people such an option would confuse them more than it would help them.

PR #843, opened 2024-08-14
feat: allow not waiting for actions

### LINKED ISSUE #833: Allow not waiting for actions
### TL;DR

In some use cases, waiting for actions to complete does not make sens, for example deleting a server can be done in a fire and forget manner to prevent waiting too long the resource are deleted.

We proposed users to put the process in the background to continue with their scripts, but this can lead to multiple problems:
- We do not have the information if the actions that we created, was actually created without error.
- We have inconsistent behavior when deleting resources in batches, the first 10 will have the expected behavior, but the rest of the resources will still have to wait for the first batch to complete.

- https://github.com/hetznercloud/cli/issues/613
- https://github.com/hetznercloud/cli/issues/489
- https://github.com/hetznercloud/cli/issues/284

### Expected behavior

Provide a --no-wait flag to bypass any actions waiting logic in our commands.

phm07:
This PR adds the `no-wait` option, which allows to skip waiting for actions.

Closes #833

Related to #613, #489, #284

### DIFF
--- a/internal/state/actions.go
+++ b/internal/state/actions.go
@@ -28,7 +28,7 @@ func (c *state) WaitForActions(cmd *cobra.Command, ctx context.Context, actions 
 			return nil
 		}
 		for _, action := range actions {
-			ui.SkipActionWaitMessage(action)
+			cmd.Println(ui.SkipActionWaitMessage(action))
 		}
 		return nil
 	}


internal/cmd/config/helptext/preferences.txt
```diff
@@ -12,6 +12,10 @@
 │ endpoint         │ Hetzner Cloud API    │ string      │ endpoint         │ HCLOUD_ENDPOINT         │ --endpoint      │
 │                  │ endpoint             │             │                  │                         │                 │
 ├──────────────────┼──────────────────────┼─────────────┼──────────────────┼─────────────────────────┼─────────────────┤
+│ no-wait          │ If true, do not wait │ boolean     │ no_wait          │ HCLOUD_NO_WAIT          │ --no-wait       │
```
apricote, line 15: I prefer `wait`.

For example, if the user set this to `true` in their config file, but want to wait for one specific call, they would need to set `--no-wait=false` which is a double negative and confusing.

internal/state/actions.go
```diff
@@ -13,6 +13,10 @@ import (
 )
 
 func (c *state) WaitForActions(cmd *cobra.Command, ctx context.Context, actions ...*hcloud.Action) error {
+	if noWait, _ := cmd.Flags().GetBool("no-wait"); noWait {
+		return nil
```
apricote, line 17: Should we log the action that we did not wait for? This way the user knows that something is still happening in the background and they could theoretically request the action status at a later time through the API.

internal/cmd/config/helptext/preferences.txt
apricote: Should we add general support for `--no-[bool-option]` flags?

internal/state/actions.go
apricote: Sounds good :+1:

internal/cmd/config/helptext/preferences.txt
jooola: Since we have multiple source of configuration, I think having a negative flag for our bool configuration is indeed handy.

For example if we have --wait, we should have --no-wait, or --include-deprecated-architecture and --no-include-deprecated-architecture

apricote: > This is something for a different PR then. So we just go for `--wait=false` now?

I would prefer that, yes.

internal/state/actions.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func (c *state) WaitForActions(cmd *cobra.Command, ctx context.Context, actions ...*hcloud.Action) error {
```
apricote, line 16: Do we have "composite" commands in the CLI, where we wait for some actions and then make more API calls? These could fail if we do not wait for the initial actions.

apricote: Sounds fine to me.

We should also look at these cases and decide if its really necessary to wait for the running actions or the next call is independent and can still be made. (ie. is it necessary for the server to start before enabling the protection?)

PR #844, opened 2024-08-14
fix(image): only accept numerical ids for update, delete, label

phm07:
Since `Image().Get()` is deprecated, we should switch to an alternative.

All the modifying calls for images are only allowed for `type=backup` and `type=snapshot`. These types cannot have names, so the user is required to pass in an ID in anyway. The easiest alternative is `Image().GetByID()`.

### DIFF
--- a/internal/cmd/image/delete.go
+++ b/internal/cmd/image/delete.go
@@ -1,6 +1,9 @@
 package image
 
 import (
+	"fmt"
+	"strconv"
+
 	"github.com/spf13/cobra"
 
 	"github.com/hetznercloud/cli/internal/cmd/base"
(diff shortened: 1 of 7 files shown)


apricote (changes requested):
I agree in principal that we should get rid of deprecated behavior.

For image labels though, this is unecessary. Users can only change the labels of images with `type=backup` and `type=snapshot`. These images always have empty names, so it does not matter that you can not uniquely identify the image to be labeled through the name. Users have to pass the ID in all cases.

We can instead modify this to use `GetByID()`.

apricote (approved):

PR #847, opened 2024-08-19
feat(certificate): allow retrying managed certificate issuance

phm07:
Somehow this API call didn't have a corresponding command in the CLI, so this PR adds it.

### DIFF
--- a/internal/cmd/certificate/retry.go
+++ b/internal/cmd/certificate/retry.go
@@ -1,0 +1,44 @@
+package certificate
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/hetznercloud/cli/internal/cmd/base"
+	"github.com/hetznercloud/cli/internal/hcapi2"
+	"github.com/hetznercloud/cli/internal/state"
+)
+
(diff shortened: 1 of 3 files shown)


internal/cmd/certificate/retry.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/hetznercloud/cli/internal/cmd/base"
+	"github.com/hetznercloud/cli/internal/hcapi2"
+	"github.com/hetznercloud/cli/internal/state"
+)
+
+var RetryCmd = base.Cmd{
+	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
+		return &cobra.Command{
+			Use:                   "retry <certificate>",
```
apricote, line 16: > It is debatable whether retry or retry-issuance is a better name for the command. retry-issuance is more precise, but also very cumbersome to use. Although it probably won't be used much anyway, since this feature was missing for 3 years and nobody noticed.

Its called `retry` in the API and `RetryIssuance` in Go, so we are not consistent any way.

I lean towards `retry` to be consistent with the API.

apricote (approved):

PR #849, opened 2024-08-22
fix(load-balancer): show message if none of --server, --label-selector or --ip is set

phm07:
Previously, if none of these flags were provided, only the error `--server, --label-selector, and --ip are mutually exclusive` would show, which is not wrong but doesn't include that exactly one is required.

We could also use `cmd.MarkFlagsOneRequired()` and `cmd.MarkFlagsMutuallyExclusive()` but the errors are not very pretty.

### DIFF
--- a/internal/cmd/util/util.go
+++ b/internal/cmd/util/util.go
@@ -88,8 +88,8 @@ func ChainRunE(fns ...func(cmd *cobra.Command, args []string) error) func(cmd *c
 	}
 }
 
-func AnySet(s string, ss ...string) bool {
-	for _, s := range append(ss, s) {
+func AnySet(ss ...string) bool {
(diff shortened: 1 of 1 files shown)


internal/cmd/util/util.go
```diff
@@ -88,6 +88,15 @@ func ChainRunE(fns ...func(cmd *cobra.Command, args []string) error) func(cmd *c
...
 }
 
+func AnySet(s string, ss ...string) bool {
+	for _, s := range append(ss, s) {
+		if s != "" {
+			return true
+		}
+	}
+	return false
+}
```
jooola, line 98: Variable naming here is really confusing, and we seem to be shadowing a variable.

apricote (approved):

PR #850, opened 2024-08-26
chore: fix build error in 0223f7d

phm07:
The CI didn't catch this error, so it ended up in main.


apricote (approved):

apricote:
Maybe `golangci-lint` only runs on changed files vs all files?

apricote:
Oh for this one I think it was just a missing rebase. Our CI only runs on the `HEAD` commit of the PR, not on the merge between `main` and `HEAD`. So if we make some changes in another PR they are not necessarily applied to existing pull requests.

PR #862, opened 2024-09-06
fix: inconsistent base for size units in description

### LINKED ISSUE #838: Use consistent size unit in server type description
### TL;DR

The `included_traffic` displayed in the server type describe command is using a different `base` for the size unit from the rest of the sizes.

### Expected behavior

Have a consistent base for the size units we display to the users.

### Observed behavior

image

### Minimal working example

```
hcloud server-type describe cpx21
```

### Log output

_No response_

### Additional information

_No response_

phm07:
This PR changes `€/TB` to `€/TiB`, which more accurately reflects how traffic is actually billed. It is now also consistent with the unit for included traffic.

Closes #838


jooola:
Did we update the docs to reflect this as well ?

apricote:
PR was closed as this change goes against the unit documented in the API docs and the frontend.

PR #870, opened 2024-09-12
refactor: prepare for e2e tests

phm07:
This PR adds refactors that are necessary for the e2e tests.
Changes:
- Subcommand initialization has been moved to `internal/cli/root.go`
- `output.Table`s now accept writers as output. This is needed to catch command output. If we only catch stdout we can only run one command at once or there will be interference.

### DIFF
--- a/internal/cmd/util/util_test.go
+++ b/internal/cmd/util/util_test.go
@@ -199,7 +199,7 @@ func TestDescribeFormat(t *testing.T) {
 		Bar: "bar",
 	}, "Foo is: {{.Foo}} Bar is: {{.Bar}}")
 
-	assert.NoError(t, err)
+	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


internal/cmd/context/create_test.go
```diff
@@ -13,6 +13,10 @@ import (
 )
 
 func TestCreate(t *testing.T) {
+	// Make sure we don't have any environment variables set that could interfere with the test
+	t.Setenv("HCLOUD_TOKEN", "")
+	t.Setenv("HCLOUD_CONTEXT", "")
```
apricote, line 18: To be on the safe side we could get the env for all variables with prefix `HCLOUD_` and unset them.

Unfortunately Go has no `os.GetEnvMap()`, and `os.Environ()` returns `[]string` with `KEY=VALUE` pairs, so lots of additional effort to parse it and I am not sure its worth it.

apricote (approved):

PR #871, opened 2024-09-16
ci: build and test with Go 1.23

phm07:
This PR makes the CI pipeline use Go 1.23 for building and testing. Only setting the `toolchain` in `go.mod` is not enough, because the [setup-go action does not support the `toolchain` directive yet](https://github.com/actions/setup-go/issues/457), we have to set the Go version to 1.23 in all setup-go CI steps manually.
We could also set the `go` version in `go.mod` to 1.23 since we do not have any dependents, but that would be misusing the `go` directive, since it is only intended to specify the minimum supported language version of the project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
 
+env:
+  GOTOOLCHAIN: local
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
 
+env:
+  GOTOOLCHAIN: local
+
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: go.mod
+          go-version: "1.23"
```
apricote, line 18: Once a new version of Go 1.23 is released, `setup-go` will install that. When the jobs start running, `go` will see the `toolchain` directive and download that version instead. This is inefficient.

Probably okay for now, can you create a tracking issue internally to go back to `go-version-file` once a fix for https://github.com/actions/setup-go/issues/457 lands.

apricote (approved):

.github/workflows/build.yml
apricote: That would be a reasonable workaround for now, but sounds very annoying to add everywhere tbh.

apricote (approved):

PR #902, opened 2024-11-08
feat: drop support for go 1.21

jooola:
Drop support for go 1.21 (we already build with go 1.23).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/hetznercloud/cli
 
-go 1.21
+go 1.22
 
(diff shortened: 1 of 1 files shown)


apricote:
Is there any syntax / stdlib that you want to use from Go 1.22?

apricote (approved):

jooola:
Replacing the mock library with the one from uber bumps the min go version to 1.22, I didn't want to do this silently.

https://github.com/hetznercloud/cli/pull/903

PR #906, opened 2024-11-08
ci: switch to releaser-pleaser

apricote:
Switch to [releaser-pleaser](https://github.com/apricote/releaser-pleaser) for planning new releases. This adds the following new features:

- We can easily cut pre-releases, these are always useful if we want to allow some customers to test features before making an official release or to validate changes made to the release CI workflows
- We can add our own release notes to each release to better inform customers about the new features we added or about any deprecations.


apricote:
After this is merged and a new PR by releaser-pleaser was opened we can close #890.

jooola (approved):

.github/workflows/releaser-pleaser.yaml
```diff
@@ -0,0 +1,24 @@
+name: releaser-pleaser
```
jooola, line 1:
```suggestion
name: Releaser-pleaser
```

PR #954, opened 2025-01-10
test: add Load Balancer Type e2e tests

phm07:


test/e2e/load_balancer_type_test.go
```diff
@@ -0,0 +1,75 @@
...
+						Int().Newline()).
+					End(),
+				out,
+			)
+		})
+
+		t.Run("json", func(t *testing.T) {
+			out, err := runCommand(t, "load-balancer-type", "list", "-o=json")
+			require.NoError(t, err)
+			assert.True(t, json.Valid([]byte(out)), "is valid JSON")
```
jooola, line 38: This isn't testing much, should we test the data inside?

test/e2e/load_balancer_type_test.go
```diff
@@ -0,0 +1,75 @@
...
+		})
+	})
+
+	t.Run("describe", func(t *testing.T) {
```
jooola, line 42: There is no JSON output for describe?

test/e2e/load_balancer_type_test.go
jooola: Load balancer type do not change every day, I am not against hardcoded values.

Or at least test the shape of the json document?

jooola: Not necessarily, we could only check for the first element.

PR #960, opened 2025-01-17
refactor: order by key when printing labels

phm07:
This PR adds sorting of labels by their keys, which is useful for testing, since iteration order over maps in Go is undefined by default.
It uses the Go 1.23 iteration feature. ~~If we want to use this, we should probably bump the Go version in a separate PR.~~


apricote (approved):

apricote:
I do not think this is worthy of being added to the Changelog, does not really matter for customers and is more about test behaviour. I updated the title for this.

PR #967, opened 2025-01-23
refactor: split up label command into separate Fetch/GetLabels/GetIDOrName

phm07:
This change is intended for an upcoming resource that has labels but doesn't have numeric IDs.

### DIFF
--- a/internal/cmd/base/labels.go
+++ b/internal/cmd/base/labels.go
@@ -14,20 +14,20 @@ import (
 )
 
 // LabelCmds allows defining commands for adding labels to resources.
-type LabelCmds struct {
+type LabelCmds[T any] struct {
 	ResourceNameSingular   string
 	ShortDescriptionAdd    string
 	ShortDescriptionRemove string
 	NameSuggestions        func(client hcapi2.Client) func() []string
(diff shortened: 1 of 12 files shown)


internal/cmd/base/labels.go
```diff
@@ -20,8 +20,10 @@ type LabelCmds struct {
 	ShortDescriptionRemove string
 	NameSuggestions        func(client hcapi2.Client) func() []string
 	LabelKeySuggestions    func(client hcapi2.Client) func(idOrName string) []string
-	FetchLabels            func(s state.State, idOrName string) (map[string]string, int64, error)
-	SetLabels              func(s state.State, id int64, labels map[string]string) error
+	Fetch                  func(s state.State, idOrName string) (any, error)
+	SetLabels              func(s state.State, resource any, labels map[string]string) error
+	GetLabels              func(resource any) map[string]string
+	GetIDOrName            func(resource any) string
```
apricote, line 26: We could make `LabelCmds` generic over the resource structs to get rid of the `any` here

apricote (commented):
Where do you intend to use this?

apricote (approved):

PR #994, opened 2025-02-28
refactor: add type for cobra completion func

apricote:
Makes the code more readable by giving a keyword to the recurring type.


jooola (approved):

apricote:
Looks like Cobra 1.9 is adding a type for this anyway: https://github.com/spf13/cobra/pull/2220/files

Will use that type instead then.

Issue #1004, opened 2025-03-10
Create Load Balancer with Network attached


geberl:
### TL;DR

For creating a LB and attaching it to a private network I currently need to run two commands (first create the LB, then do an attach). The API would allow for attaching the LB to the Network right on create.

### Expected behavior

The same is already possible for servers (`--network` with id or name).

Also it looks like `hcloud-go` already includes a `*Network` in `type LoadBalancerCreateOpts`: https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2/hcloud?utm_source=godoc#LoadBalancerCreateOpts

Can we get the `--network` flag for Load Balancers as well?

PR #1013, opened 2025-03-20
feat(load-balancer): allow specifying network on create

phm07:
This PR adds the ability to automatically attach a Load Balancer to a network after creation using the `--network` flag.

See #1004

### DIFF
--- a/internal/cmd/loadbalancer/create.go
+++ b/internal/cmd/loadbalancer/create.go
@@ -45,7 +45,7 @@ var CreateCmd = base.CreateCmd{
 		cmd.Flags().StringSlice("enable-protection", []string{}, "Enable protection (delete) (default: none)")
 		_ = cmd.RegisterFlagCompletionFunc("enable-protection", cmpl.SuggestCandidates("delete"))
 
-		cmd.Flags().String("network", "", "Name or ID of the network the Load Balancer should be attached to on creation")
+		cmd.Flags().String("network", "", "Name or ID of the Network the Load Balancer should be attached to on creation")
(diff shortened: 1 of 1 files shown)


internal/cmd/loadbalancer/create.go
```diff
@@ -43,6 +43,9 @@ var CreateCmd = base.CreateCmd{
 		cmd.Flags().StringSlice("enable-protection", []string{}, "Enable protection (delete) (default: none)")
 		_ = cmd.RegisterFlagCompletionFunc("enable-protection", cmpl.SuggestCandidates("delete"))
 
+		cmd.Flags().Int64("network", 0, "ID of the network the Load Balancer should be attached to on creation")
+		_ = cmd.RegisterFlagCompletionFunc("network", cmpl.SuggestCandidatesF(client.Network().IDs))
```
jooola, line 47: Should we support both ID or name?

internal/cmd/loadbalancer/create.go
```diff
@@ -43,6 +45,9 @@ var CreateCmd = base.CreateCmd{
 		cmd.Flags().StringSlice("enable-protection", []string{}, "Enable protection (delete) (default: none)")
 		_ = cmd.RegisterFlagCompletionFunc("enable-protection", cmpl.SuggestCandidates("delete"))
 
+		cmd.Flags().String("network", "", "Name or ID of the network the Load Balancer should be attached to on creation")
```
jooola, line 48:
```suggestion
		cmd.Flags().String("network", "", "Name or ID of the Network the Load Balancer should be attached to on creation")
```

jooola (approved):

Issue #1016, opened 2025-03-27
docs: Publish --help pages in the repository


apricote:
### TL;DR

Right now the help output is only available through the CLI itself. For easier discoverability it should also be published as Markdown in the repository.

### Expected behavior

- Automatically build the help pages from the cobra CLI and output as Markdown to `docs/reference/`
- Link to this in our README
- Add CI step to validate that the help pages are up-to-date

As a reference, I did this before in https://github.com/apricote/hcloud-upload-image/pull/46/files

PR #1018, opened 2025-03-27
docs: automatically generate command help documentation

phm07:
This PR adds auto-generated documentation for commands like described in https://github.com/hetznercloud/cli/issues/1016

### DIFF
--- a/docs/reference/hcloud.md
+++ b/docs/reference/hcloud.md
@@ -22,23 +22,23 @@ A command-line interface for Hetzner Cloud
 ### SEE ALSO
 
 * [hcloud all](hcloud_all.md)	 - Commands that apply to all resources
-* [hcloud certificate](hcloud_certificate.md)	 - Manage certificates
+* [hcloud certificate](hcloud_certificate.md)	 - Manage Certificates
 * [hcloud completion](hcloud_completion.md)	 - Output shell completion code for the specified shell
 * [hcloud config](hcloud_config.md)	 - Manage configuration
 * [hcloud context](hcloud_context.md)	 - Manage contexts
-* [hcloud datacenter](hcloud_datacenter.md)	 - Manage datacenters
+* [hcloud datacenter](hcloud_datacenter.md)	 - View Datacenters
 * [hcloud firewall](hcloud_firewall.md)	 - Manage Firewalls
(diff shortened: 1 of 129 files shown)


README.md
```diff
@@ -58,6 +58,8 @@ how to embed it yourself.
 
 See `hcloud help` for a list of commands.
 
+You can also check out the [command reference](docs/reference) for help.
```
jooola, line 61: Could we use `docs/references` plural please ?

jooola (commented):
Before this PR, could you build a small docs structure to guide the readers by the hand with a set of README's, similar to how the fleeting docs is structured: 

https://gitlab.com/hetznercloud/fleeting-plugin-hetzner/-/tree/main/docs?ref_type=heads

This is to follow the https://diataxis.fr/ pattern.

docs/reference/hcloud.md
```diff
@@ -0,0 +1,44 @@
...
+Hetzner Cloud CLI
+
+### Synopsis
+
+A command-line interface for Hetzner Cloud
+
+### Options
+
+```
+      --config string            Config file path (default "/Users/paul/.config/hcloud/cli.toml")
```
apricote, line 12: Hehe, you might want to fix this, sounds like CI will always error.

apricote (changes requested):

README.md
jooola: We can follow https://diataxis.fr/reference/ to the letter and use the singlar form, but in that case I'll need to update the fleeting plugin.

jooola:
Should we wait for #1021 before merging this?

jooola:
See https://github.com/hetznercloud/cli/pull/1018#pullrequestreview-2720949815

And if you are referring to the generated docs files, I am not sure, going down sub directories to search for a specific feature can be annoying, if everything is in a single directory, one has a full overview of what is available.

We also already have some table of content in each command group.

apricote:
> Right now it is not really easy to search through documentation because everything is on the top level. Should we introduce a folder structure?

Agreeing with @jooola, everything is already linked, so a folder structure does not really add a lot IMO.

---

> Before this PR, could you build a small docs structure to guide the readers by the hand with a set of README's, similar to how the fleeting docs is structured:

I think we should add this, but not necessarily before this PR. Happy to see this merged and the overview added to the README right after.

apricote (approved):

PR #1025, opened 2025-04-03
fix: config option flags sometimes not parsed correctly

### LINKED ISSUE #845: fix: --debug is sometimes not applied
### TL;DR

`--debug` is sometimes not applied when passed in late in the arguments.

### Expected behavior

`--debug` should work, no matter at what position in the argument list it is being passed.

### Observed behavior

On some subcommands, `--debug` needs to be passed early to work. I did not find a pattern to this.

- :heavy_check_mark: `hcloud --debug server attach-to-network --network foo bar`
- :heavy_check_mark: `hcloud server --debug attach-to-network --network foo bar`
- :heavy_check_mark: `hcloud server attach-to-network --debug --network foo bar`
- :x: `hcloud server attach-to-network --network foo --debug bar`
- :x: `hcloud server attach-to-network --network foo bar --debug`

### Minimal working example

_No response_

### Log output

_No response_

### Additional information

```
$ hcloud version
hcloud 1.45.0-dev
```

phm07:
See https://github.com/hetznercloud/cli/issues/845 for a description of the error

While parsing the config's `FlagSet`, if the parser reached a flag that was not a config option, it would terminate early and return an unknown flag error. Since the error was ignored silently, this behavior was not apparent to the user.

This PR fixes this behavior by using the `ParseErrorsWhitelist` to allow unknown flags in the config's `FlagSet`. This also means we can now return any other parsing errors properly instead of ignoring them.

Fixes #845

### DIFF
--- a/internal/state/config/config_test.go
+++ b/internal/state/config/config_test.go
@@ -1,0 +1,48 @@
+package config_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/hetznercloud/cli/internal/cli"
+	"github.com/hetznercloud/cli/internal/state/config"
+	"github.com/hetznercloud/cli/internal/testutil"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+)
+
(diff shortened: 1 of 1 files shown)


jooola (commented):
Could we have a test to ensure the behavior is correct?

jooola (approved):

PR #1034, opened 2025-04-17
docs(reference): move manpages to subfolder and add README

phm07:


internal/scripts/generate_docs.go
```diff
@@ -19,15 +19,26 @@ func run() error {
...
 	}
-
-	// Create directory again
-	if err := os.MkdirAll(dir, 0755); err != nil {
-		return fmt.Errorf("error creating docs directory: %w", err)
+	for _, f := range files {
+		if !f.IsDir() && f.Name() == "README.md" {
+			// we want to keep the README.md file
+			continue
+		}
```
apricote, line 36: Do we plan to add any other reference documentation in the future? If yes, we could move the autogenerated man pages to a subdirectory and only wipe that instead of keeping a "keep-list" of all other files.

docs/reference/README.md
```diff
@@ -0,0 +1,5 @@
+# References
+
+In this directory you will find the reference documentation for the hcloud CLI.
+
+- [Command reference](manual/hcloud.md)
```
jooola, line 5:
```suggestion
- [Command manual](manual/hcloud.md)
```

Less "reference" repeating, and more details about what's inside?

jooola:
Aside from the little nit, looks good.

Could you update the title of the PR to reflect the changes?

jooola (approved):

PR #1053, opened 2025-05-22
ci: use mise for tool management

phm07:
This PR introduces [mise](https://github.com/jdx/mise) to manage and install tools for our dev environments and CI. 
It cleans up our messy tool situation, where we previously had `/tools.go`, `/tools/go.mod` and some manually installed tools as well. It also replaces GitHub actions used for installing tools. Since the action is cached, installing Go + all needed tools only takes 3 seconds, where it before took 9 seconds just to install Go alone, which is another nice benefit.


apricote:
Works after yeeting `mockgen` from `~/go/bin/mockgen`. Unfortunate that mise does not override this.

apricote:
It keeps the versions "locally" and appends the `path` through `eval $(mise activate zsh)` (can be run automatically with direnv or mise hooks for zsh|bash|fish|nushell).

In `cli` this branch:

```
$ which goreleaser
/home/julian/.local/share/mise/installs/goreleaser/v2.9.0/goreleaser
```

I think the issue for this bug may be related to my local `$PATH` setup where `~/go/bin` is currently before the `mise` paths.

apricote:
@phm07 and I took another look at this. Turns out it's Jetbrains Goland that adds the duplicate `~/go/bin` to my `$PATH` at the wrong location.

I was able to get this fixed by disabling `Tools > Terminal > Shell integration`, which stops the Goland integrated terminal from adding more things to the PATH. `mise` now works great!

apricote (approved):

PR #1080, opened 2025-06-16
feat: improve error formatting

### LINKED ISSUE #1075: Improve invalid input error handling and printing
### TL;DR

We currently print the invalid input errors the same way we print all other errors. 

The invalid input error holds additional details about the reasons why the invalid input error occurred, this data is not shown to the users. 

```
$ hcloud server create --name "a a" --image debian-12 --type cpx11
hcloud: invalid input in field 'name' (invalid_input, 6ea253683232ea88)
```

### Expected behavior

Handle invalid input errors and print all the details for the users to fully understand why the invalid input error failed. Below is an example of invalid input error with the details:

```
{
  "error": {
    "message": "invalid input in field 'name'",
    "code": "invalid_input",
    "details": {
      "fields": [
        {
          "name": "name",
          "messages": [
            "Name must be a valid hostname."
          ]
        }
      ]
    }
  }
}
```

phm07:
This PR adds additional error handling and formatting for `invalid_input` errors. See #1075 for more details.

Before:
```
$ hcloud server create --name "a a" --image debian-12 --type cpx11
hcloud: invalid input in field 'name' (invalid_input, ca050180d7ec9473)
```

After:
```
$ hcloud server create --name "a a" --image debian-12 --type cpx11 
hcloud: invalid input in field 'name' (invalid_input, ca050180d7ec9473)
- name: Name must be a valid hostname.
```

~~Note: As we are not inside the `hcloud-go` package we can't access the internal Correlation ID. Since invalid input errors are user errors we probably don't need them anyway.~~

Fixes #1075

### DIFF
--- a/internal/cmd/util/util_test.go
+++ b/internal/cmd/util/util_test.go
@@ -505,7 +505,7 @@ func TestFormatHcloudError(t *testing.T) {
 			},
 		},
 	}
-	assert.Equal(t, `Invalid input
+	assert.Equal(t, `Invalid input (invalid_input)
 - foo: Invalid value
 - bar: Must be a number, Must be greater than 0`, util.FormatHcloudError(invalidInputError))
 }


internal/cmd/util/util.go
```diff
@@ -377,3 +378,29 @@ func IterateInOrder[M ~map[K]V, K cmp.Ordered, V any](m M) iter.Seq2[K, V] {
...
+	var hcloudErr hcloud.Error
+	if !errors.As(err, &hcloudErr) {
+		return err.Error()
+	}
+
+	switch hcloudErr.Code {
+	case hcloud.ErrorCodeInvalidInput:
+		details := hcloudErr.Details.(hcloud.ErrorDetailsInvalidInput)
+		var errBuilder strings.Builder
+		errBuilder.WriteString(hcloudErr.Message)
```
apricote, line 396: If we want to include the correlation ID we can still use `hcloudErr.Error()` here and add the details at the end.

```suggestion
		errBuilder.WriteString(hcloudErr.Error())
```

PR #1082, opened 2025-06-17
feat(firewall): display applied to resources when using label selector

phm07:
This PR adds the ability to see which resources a Firewall is applied to when using a label selector.

Can only be merged once https://github.com/hetznercloud/hcloud-go/pull/667 is merged and released

### DIFF
--- a/internal/cmd/firewall/describe_test.go
+++ b/internal/cmd/firewall/describe_test.go
@@ -70,6 +70,12 @@ func TestDescribe(t *testing.T) {
 		Get(gomock.Any(), "test").
 		Return(fw, nil, nil)
 	fx.Client.ServerClient.EXPECT().
+		ServerName(int64(123)).
+		Return("appliedServer1")
+	fx.Client.ServerClient.EXPECT().
+		ServerName(int64(456)).
+		Return("appliedServer2")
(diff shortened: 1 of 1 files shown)


internal/cmd/firewall/describe.go
```diff
@@ -79,6 +79,15 @@ var DescribeCmd = base.DescribeCmd[*hcloud.Firewall]{
 					cmd.Printf("    Server Name:\t%s\n", s.Client().Server().ServerName(resource.Server.ID))
 				case hcloud.FirewallResourceTypeLabelSelector:
 					cmd.Printf("    Label Selector:\t%s\n", resource.LabelSelector.Selector)
+					if len(resource.AppliedToResources) > 0 {
+						cmd.Printf("    Applied to resources:\n")
+						for _, appliedTo := range resource.AppliedToResources {
+							switch appliedTo.Type {
+							case hcloud.FirewallResourceTypeServer:
+								cmd.Printf("      Server %d\n", appliedTo.Server.ID)
+							}
```
apricote, line 88: This could reuse the same printing logic as top level server resources. That would make it consistent and also print the name for the servers.

PR #1090, opened 2025-07-01
docs: remove description option from firewall delete-rule

mkreis:
--description is not a valid option of delete-rule


apricote:
Hey @mkreis,

I think `--description` is a valid flag for `hcloud firewall delete-rule`: https://github.com/hetznercloud/cli/blob/056955cfd1dcfa2c3a9ad9183e105451f73bb4be/internal/cmd/firewall/delete_rule.go#L39

Can you explain why you do not think that it is a valid option?

mkreis:
@apricote My bad, I tried to delete a rule using its description and got an error message. It seems like the error was on my side -- sorry

PR #1164, opened 2025-09-22
feat(server,load-balancer): add --ip-range argument to attach-to-network commands

phm07:
A new `ip_range` attribute was introduced into the API ([Changelog](https://docs.hetzner.cloud/changelog#2025-09-11-attach-to-network-support-ip-range)). This PR implements new arguments to allow setting this property when attaching Servers or Load Balancers to a Network.

### DIFF
--- a/docs/reference/manual/hcloud_load-balancer_attach-to-network.md
+++ b/docs/reference/manual/hcloud_load-balancer_attach-to-network.md
@@ -11,7 +11,7 @@ hcloud load-balancer attach-to-network [--ip <ip>] --network <network> <load-bal
 ```
   -h, --help             help for attach-to-network
       --ip ip            IP address to assign to the Load Balancer (auto-assigned if omitted)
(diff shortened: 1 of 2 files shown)


internal/cmd/loadbalancer/attach_to_network.go
```diff
@@ -27,6 +28,7 @@ var AttachToNetworkCmd = base.Cmd{
 		_ = cmd.MarkFlagRequired("network")
 
 		cmd.Flags().IP("ip", nil, "IP address to assign to the Load Balancer (auto-assigned if omitted)")
+		cmd.Flags().IPNet("ip-range", net.IPNet{}, "IP range in CIDR block notation of the subnet to attach to")
```
lukasmetzner, line 31: We may want to update both docstrings. 

`ip-range` would also fit the `(auto-assigned if omitted)` case, if `ip` is not provided.

In the API you can specify one of them, both, or none.

PR #1197, opened 2025-10-16
fix: more readable default time format

apricote:
Changes the time format from `time.UnixDate` to `time.DateTime + timezone`.

- Old: `Mon Jan 10 15:04:05 MST 2006`
- New: `2006-01-10 15:04:05 MST`

I personally find this much easier to read.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -65,7 +65,7 @@ func (r RegexBuilder) SeparatedByWhitespace(lits ...string) RegexBuilder {
 
 func (r RegexBuilder) Datetime() RegexBuilder {
 	// 2006-01-02 15:04:05 MST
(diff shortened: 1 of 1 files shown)


apricote:
>  Is this a breaking change?

In theory yes, as is every new property we add to `describe` or `list` outputs.

I hope no one was parsing the old format. 

Maybe we should make it a global option and change the default? Then anyone can move back to the old format if they want to.

phm07:
I think a date format option would be a little overkill. It is probably safe to assume that users use JSON or YAML formats when working with raw data.

jooola:
I think we missed a rebase for this PR before merging:  https://github.com/hetznercloud/cli/actions/runs/18940519844/job/54077878135#step:4:819

PR #1198, opened 2025-10-17
chore: reduce diff by hiding generated documentation by default

jooola:
I would like to reduce the amount of files shown in PRs, this hides the generated manuals by default.

Github was complaining for the Storage Boxes PR being too big.


apricote:
Even with marking the files as generated GH only shows the single-file view for the Storage Box PR. :(

(Still a valid change, not asking for it to be reverted.)

PR #1232, opened 2025-11-07
deps: fix renovate picking up mise.toml

phm07:
Currently, `goreleaser` is not supported as a default registry tool short name by renovate, so we have to annotate it manually. See https://docs.renovatebot.com/modules/manager/mise/#supported-default-registry-tool-short-names

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,10 +1,10 @@
 [tools]
 go = "1.25.4"
-"go:go.uber.org/mock/mockgen" = "v0.6.0"  # renovate: datasource=github-releases depName=uber-go/mock
-golangci-lint = "2.6.1"
-goreleaser = "v2.9.0"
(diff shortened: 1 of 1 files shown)


mise.toml
```diff
@@ -2,7 +2,7 @@
 go = "1.25.4"
 "go:go.uber.org/mock/mockgen" = "v0.6.0"  # renovate: datasource=github-releases depName=uber-go/mock
 golangci-lint = "2.6.1"
-goreleaser = "v2.9.0"
+goreleaser = "v2.9.0"                     # renovate: datasource=github-releases depName=goreleaser/goreleaser
```
jooola: What about being explicit about the backend and specify the full path to the project?


```suggestion
ubi:goreleaser/goreleaser = "v2.9.0"                     # renovate: datasource=github-releases depName=goreleaser/goreleaser
```

I don't know if this is using ubi or go.

(I would even do the same for golangci-lint)

PR #1268, opened 2025-12-04
refactor: add base command for change protection logic

phm07:
Moving change-protection logic to a base command removes duplication and allows us to make changes more easily in the future.

### DIFF
--- a/internal/cmd/base/change_protection.go
+++ b/internal/cmd/base/change_protection.go
@@ -3,7 +3,6 @@ package base
 import (
 	"fmt"
 	"log"
-	"reflect"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/cmd/base/change_protection.go
```diff
@@ -0,0 +1,228 @@
...
+
+// ChangeProtectionCmds allows defining commands for changing a resource's protection.
+type ChangeProtectionCmds[T, Opts any] struct {
+	ResourceNameSingular    string // e.g. "Server"
+	ShortEnableDescription  string
+	ShortDisableDescription string
+	NameSuggestions         func(client hcapi2.Client) func() []string
+	AdditionalFlags         func(*cobra.Command)
+	// Fetch is called to fetch the resource to describe.
+	// The first returned interface is the resource itself as a hcloud struct, the second is the schema for the resource.
```
apricote: It feels like this comment is no longer accurate, as it describes types for `interface{}` return values, which are now strongly typed.

internal/cmd/base/change_protection.go
```diff
@@ -0,0 +1,228 @@
...
+	if cpc.FetchWithArgs != nil {
+		resource, _, err = cpc.FetchWithArgs(s, cmd, args)
+	} else {
+		resource, _, err = cpc.Fetch(s, cmd, args[0])
+	}
+	if err != nil {
+		return err
+	}
+
+	// resource is an interface that always has a type, so the interface is never nil
```
apricote: Is it actually an `interface`? That word has a specific meaning in Go

apricote (approved):

internal/cmd/base/change_protection.go
phm07: Yes, I think we can yeet the comment

PR #1278, opened 2025-12-18
feat(server,primary-ip): deprecate datacenter

phm07:
The `datacenter` attribute is deprecated in the `Primary IPs` and `Servers` API resources and will be removed after 1 July 2026. See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters for more details.

The `location` attribute already exists for Servers, and was added for Primary IPs.

Therefor the `datacenter` attribute is deprecated in favor of the `location` attribute in the following commands:

- `hcloud server create`
- `hcloud server describe`
- `hcloud server list`
- `hcloud primary-ip create`
- `hcloud primary-ip describe`
- `hcloud primary-ip list`

In list commands, the `datacenter` columns is not longer shown by default, and is replaced with the `location` column.

### DIFF
--- a/internal/cmd/server/create.go
+++ b/internal/cmd/server/create.go
@@ -444,8 +444,7 @@ func createOptsFromFlags(
 		// Else -> Backwards-compatible split
 		var datacenterID int64
 		datacenterID, err = strconv.ParseInt(datacenterIDOrName, 10, 64)
-
-		if err != nil {
+		if err == nil {
 			// Input was a valid number/ID
 			var datacenter *hcloud.Datacenter
 			datacenter, _, err = s.Client().Datacenter().GetByID(s, datacenterID)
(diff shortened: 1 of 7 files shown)


internal/cmd/server/create.go
```diff
@@ -41,6 +42,10 @@ var CreateCmd = base.CreateCmd[*createResult]{
 		cmd := &cobra.Command{
 			Use:   "create [options] --name <name> --type <server-type> --image <image>",
 			Short: "Create a Server",
+			Long: `Create a Server.
+
+The --datacenter flag is deprecated. Use --location instead.
+See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters`,
```
jooola:
```suggestion
See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters.`,
```

PR #1300, opened 2026-01-08
feat(table): mark columns as deprecated and show warning

### LINKED ISSUE #1295: feat(table): mark columns as deprecated and show warning
### TL;DR

Add a method to mark columns in tables as deprecated and show a warning if the column is used.

### Expected behavior

It should be possible to mark a column as deprecated with a message containing details. Whenever the user selects to show that column they should see the deprecation message on `stderr`.

Code example to mark the column deprecated:

```go
var ListCmd = &base.ListCmd[*hcloud.Server, schema.Server]{
	OutputTable: func(t *output.Table[*hcloud.Server], client hcapi2.Client) {
		t.
			AddAllowedFields(&hcloud.Server{}).
			AddFieldFn("datacenter", func(server *hcloud.Server) string {
				if server.Datacenter != nil {
					return server.Datacenter.Name
				}
				return "-"

			}).
			MarkFieldAsDeprecated("datacenter", "The datacenter column is deprecated. Use location column instead.")
	}
}
```

And then the user sees the warning message when the use the column:

```shell
$ hcloud server list -o=columns=id,datacenter
Warning: The datacenter column is deprecated. Use location column instead. # stderr
ID    DATACENTER
123   hel1-dc2
```

phm07:
Closes #1295

### DIFF
--- a/internal/cmd/base/list.go
+++ b/internal/cmd/base/list.go
@@ -164,10 +164,8 @@ func (lc *ListCmd[T, S]) Run(s state.State, cmd *cobra.Command, args []string) e
 	t := output.NewTable[T](out)
 	lc.OutputTable(t, s.Client())
 
-	warnings, err := t.ValidateColumns(cols)
-	if err != nil {
-		return err
-	}
+	warnings, _ := t.ValidateColumns(cols)
+	// invalid columns are already checked in output.validateOutputFlag(), we only need the warnings here
 	for _, warning := range warnings {
 		cmd.PrintErrln("Warning:", warning)
 	}


internal/cmd/base/list.go
```diff
@@ -163,6 +163,15 @@ func (lc *ListCmd[T, S]) Run(s state.State, cmd *cobra.Command, args []string) e
 
 	t := output.NewTable[T](out)
 	lc.OutputTable(t, s.Client())
+
+	warnings, err := t.ValidateColumns(cols)
+	if err != nil {
+		return err
```
phm07: This check was missing before. Since invalid columns were silently ignored before and now throw an error, this would be a breaking change. Do we want to keep it anyway?

jooola: I usually prefer not to maintain bug compatibility.

But in this case, maybe we can log a warning?

PR #1324, opened 2026-02-01
feat: added options for static-ip & custom root CA

normanre:
Hello everyone,

I added these two options to configure a http.Client for the cli:

1. static-ip: With this option a static ip is used and DNS resolution is skipped but still the host of endpoint is used
2. root-ca: With this option someone can define a custom root ca

I did this because DNS and certificates resolution does not work as expected on Termux for Android. With these two additions this cli can be used for Android.

One quirk with Termux is that `export HCLOUD_CONFIG="$HOME/.config/hcloud/cli.toml"` has to be set (for example in .bashrc)

I hope this helps also someone else.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -92,13 +92,14 @@ func (c *state) newClient() (hcapi2.Client, error) {
 		return nil, err
 	}
 
+	var debugWriter io.Writer
+
 	if debug {
 		filePath, err := config.OptionDebugFile.Get(c.config)
 		if err != nil {
 			return nil, err
 		}
 
-		var debugWriter io.Writer
 		if filePath == "" {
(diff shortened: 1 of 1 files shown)


phm07:
Where did you install `hcloud` from?

Termux already offers an unofficial hcloud build that should work with their distro:
https://github.com/termux/termux-packages/blob/master/packages/hcloud/build.sh
You should be able to install it using `pkg install hcloud`.

Their build does not set `CGO_ENABLED=0` as this apparently causes DNS issues in termux, see https://github.com/termux/termux-packages/issues/14332.

normanre:
OMG. I just installed the binary straight. Sorry, for the disturbance.

PR #1389, opened 2026-04-23
Add option to use proxy

1fexd:
Adds an option to the CLI that allows specifying a proxy to use; Requires https://github.com/hetznercloud/hcloud-go/pull/838

### DIFF
--- a/internal/state/config/options.go
+++ b/internal/state/config/options.go
@@ -169,6 +169,15 @@ var (
 		nil,
 	)
 
+	OptionProxy = newOpt(
+		"proxy",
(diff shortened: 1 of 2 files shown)


jooola:
I would prefer not to add a custom way of configuring a HTTP proxy in the CLI.

The default go http transport already loads proxies from the environment:
- https://pkg.go.dev/net/http?utm_source=godoc#DefaultTransport
- https://pkg.go.dev/net/http?utm_source=godoc#ProxyFromEnvironment

Could you try to use this method instead?

PR #1393, opened 2026-04-24
refactor: allow returning multiple actions on delete

phm07:
This PR allows returning multiple actions during delete.

### DIFF
--- a/internal/cmd/base/delete.go
+++ b/internal/cmd/base/delete.go
@@ -24,7 +24,7 @@ type DeleteCmd[T any] struct {
 	NameSuggestions      func(client hcapi2.Client) func() []string
 	AdditionalFlags      func(*cobra.Command)
 	Fetch                FetchFunc[T]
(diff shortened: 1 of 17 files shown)


internal/cmd/base/delete.go
```diff
@@ -115,7 +115,7 @@ func (dc *DeleteCmd[T]) Run(s state.State, cmd *cobra.Command, args []string) er
 
 	for batch := range slices.Chunk(toDelete, deleteBatchSize) {
 		results := make([]util.ResourceState, len(batch))
-		actions := make([]*hcloud.Action, 0, len(batch))
+		var actions []*hcloud.Action
```
jooola, line 118: Why change this?

jooola (approved):

internal/cmd/base/delete.go
phm07: Since the amount of actions does not necessarily equal the batch size (since some delete commands return multiple actions or none at all), there is no need to initialize the actions slice with a capacity.

PR #1405, opened 2026-05-08
feat: allow fetching token from shell command

### LINKED ISSUE #808: Store token in a more secure manner , support pairs of tokens
### TL;DR

hcloud stores the tokens in plaintext in ~/.config/hcloud/cli.toml , where any other process, including any which reads from or includes files from random paths, making it too easy to steal the tokens. 

Since hcloud enforces the use of these contexts stored in that file, and does not allow to pass the token per command (like, e.g. terraform/opentofu does)  to the process without storing it, this is highly insecure and puts the tokens into high risk of beeing stolen, rendering the 2FA for the login to the Hetzner web console effectively useless.

### Expected behavior

More secure handling of tokens, e.g.

- ask for token during cli run without storing it
- instead of storing the token in a file, store a command which is run and expected to output the token in stdout, thus allowing the usual Linux methods to store secrets in Wallets like Gnome Keyring or Keepass.
- allow handling of a token pair of a read only and read/write token, where, e.g. the read only token with lower risk can be stored in plaintext in ~/.config/hcloud/cli.toml for fast and convenient read-only commands like hcloud server list, where aggressive commands like server create or server delete automatically choose the read/write token, which might be retrieved from a wallet then.

phm07:
This PR allows the CLI to run a shell command to fetch the HCLOUD_TOKEN for each run, making it possible to use a keyring or password manager to securely store the token.

Fixes #808


internal/state/config/token.go
```diff
@@ -0,0 +1,51 @@
...
+	if tok, ok := cmdCache[cmdStr]; ok {
+		return tok, nil
+	}
+
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("cmd", "/C", cmdStr)
+	} else {
+		cmd = exec.Command("sh", "-c", cmdStr)
+	}
```
jooola, line 44: A few questions:
- How can one define this config globally and pick the right context? See point below.
- Why not pass down more details about the current context? E.g. as environment variables?
- Why use a shell at all? I'd leave this for the users to pick a shell if needed
- Could we store the command as list of strings, instead of a string? This makes passing down the arguments cleaner. Or we implement a similar approach like the docker CMD: If we have a list of string, no shell, if we have a string command, wrap it inside a shell?

internal/state/config/token.go
```diff
@@ -0,0 +1,51 @@
...
+	out, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("could not retrieve token: %w", err)
+	}
```
jooola, line 52: Should we forward the command stderr to the cli stderr?

internal/state/config/token.go
phm07: > How can one define this config globally and pick the right context? See point below.
> Why not pass down more details about the current context? E.g. as environment variables?

Good point, env variables would fix this

> Why use a shell at all? I'd leave this for the users to pick a shell if needed

A shell offers more features like pipes, input/output file redirects etc. Also it allows passing the command as one string instead of a string slice, see below.

> Could we store the command as list of strings, instead of a string? This makes passing down the arguments cleaner. Or we implement a similar approach like the docker CMD: If we have a list of string, no shell, if we have a string command, wrap it inside a shell?

List of strings would be fine, except the only problem would be how to configure it using the `hcloud context create` command. Passing it as a slice would mean passing `--token-command` multiple times which would be very cumbersome or to separate the parts with commas which would mean you couldn't use a comma in your token command. Accepting both a string and a string slice would be fine but I don't know if that would play well with config parsing.

internal/state/config/token.go
phm07: Good idea

PR #1412, opened 2026-05-22
fix(cmd): Change tense for deletion of active context

nachtjasmin:
"You are deleting" implies that I can abort the action somehow. In other user interfaces, it's common to end sentences like this with "Are you sure?". Given that the hcloud CLI doesn't do this and happily deletes the context, I changed the wording to indicate that the action was executed.


phm07 (approved):

phm07:
Thanks for the contribution 🙂

PR #1435, opened 2026-06-15
feat(server): add screenshot command

### LINKED ISSUE #1420: Add a server-side console screenshot endpoint / `hcloud server screenshot` (PNG of the VGA console, no interactive VNC session)
## Summary

I'd like to request a **server-side console screenshot** capability: an action that captures the current VGA/console framebuffer of a Cloud server and returns it as an image (PNG/JPEG), without requiring an interactive noVNC session. Ideally surfaced in the CLI as:

```
hcloud server screenshot <server> -o console.png
```

I realize a still-image of the console is a new *platform/API* capability, and that this repo is a thin wrapper over the public API — so this is filed here mainly to track the CLI surface and gauge interest. If the maintainers would rather I route the platform side through the in-Console **Feedback** menu (the channel announced on [docs.hetzner.cloud/whats-new](https://docs.hetzner.cloud/whats-new) on 2025-08-26 for product/API feature requests), I'm happy to cross-post there — a pointer would be appreciated.

## Use case

The motivating scenario is debugging a server that is **unbootable or otherwise unreachable**, where you have no other window into what's happening:

- A server stuck in the **UEFI/EDK2 shell** or boot menu.
- A **stage-1 / initramfs boot hang**, a kernel panic, or a GRUB/BIOS screen frozen before networking comes up.
- Anything where **SSH never comes up** and there is no serial-console-log API to read instead.
- **CI / automation**, where there is no human available to open the noVNC console and look at the screen. A pipeline that provisions a server, waits, and finds it unreachable currently has no programmatic way to capture "what is on the screen right now" for triage or for attaching to a failure report.

In all of these, the one piece of information that would unblock diagnosis is a single image of the console. Today there is no way to get that without a human and a browser.

## The gap

The Cloud API's only console capability is **interactive-only**. The single console endpoint is:

```
POST /v1/servers/{id}/actions/request_console
```

> "Requests credentials for remote access via VNC over websocket to keyboard, monitor, and mouse for a Server. The provided URL is valid for 1 minute […]"

It returns a `wss_url` (e.g. `wss://console.hetzner.cloud/?server_id=…&token=…`, token valid ~1 minute) plus a VNC `password`. That is a **live noVNC/VNC-over-WebSocket interactive session** — keyboard/monitor/mouse — **not a still image**.

Checked against the authoritative OpenAPI spec ([docs.hetzner.cloud/cloud.spec.json](https://docs.hetzner.cloud/cloud.spec.json)): of the 24 `/servers/{id}/actions/*` paths, `request_console` is the **only** console-related one. The canonical clients mirror this exactly — `hcloud-go` exposes only `ServerClient.RequestConsole` ("requests a WebSocket VNC console") and `hcloud-python` only `request_console` (→ `wss_url` + `password`). **Neither has any screenshot method**, because the API exposes none.

## Industry parity

Server-side console screenshots are a fairly common primitive among major providers — usually built exactly for the "instance won't boot / can't SSH in" case:

| Provider | Screenshot API? | API / CLI | Returns |
|---|---|---|---|
| **AWS EC2** | Yes | `GetConsoleScreenshot` (`aws ec2 get-console-screenshot`) | Base64-encoded **JPEG** (`imageData`), server-side, no session |
| **Google Cloud (GCE)** | Yes | `instances.getScreenshot` (`gcloud compute instances get-screenshot`) | Base64-encoded **JPEG** (`contents`); needs a virtual display device |
| **Azure VMs** | Yes | `RetrieveBootDiagnosticsData` (Boot Diagnostics) | Short-lived SAS URL to `vm.screenshot.bmp` (**BMP**) |
| **Hetzner Cloud** | **No** | `request_console` only | Interactive VNC `wss_url` + `password` — no image |
| DigitalOcean | No | — | Interactive browser/VNC console only |
| Vultr | No | — | Interactive noVNC console only |
| Linode (Akamai) | No | — | Glish (interactive VNC) / Lish (shell) only |

So this isn't universal — DigitalOcean, Vultr, and Linode also lack it — but the three hyperscalers (AWS, GCP, Azure) all provide it, and AWS/GCP in particular return it as a simple base64 image, which is a nice shape to mirror.

## Current workaround (and why it's fragile)

Because there's no native endpoint, the only way to get a console image today is to **call `request_console`, dial the `wss://` proxy, speak the RFB/VNC protocol yourself, and decode the framebuffer** — all within the ~1-minute token window. People have repeatedly built tooling to do exactly this, which is itself evidence of the demand:

- **[agentydragon/ducktape — `hetzner_vnc_screenshot`](https://github.com/agentydragon/ducktape/blob/devel/cluster/skills/hetzner_vnc_screenshot/SKILL.md)** — agent skill that does `request_console` → WebSocket → RFB (via `asyncvnc`) → PNG (Pillow). Explicitly framed as the workaround for diagnosing boot problems, kernel panics, and stuck GRUB/BIOS/UEFI screens.
- **[t-unix/hcloud-console](https://github.com/t-unix/hcloud-console)** — Go CLI that decodes the RFB 3.8 framebuffer; `--once <id>` captures a single frame and exits (scriptable). "Exists precisely because the Hetzner API exposes no native console screenshot."
- **[hilbix/hcloud-console](https://github.com/hilbix/hcloud-console)** — Python+noVNC middleware bridging the request-console VNC WebSocket to a browser viewer; canonical example of the fragile wss→RFB bridge built around the missing endpoint.
- **[RavuAlHemio — hetzner-vnc-alt-sysrq.js](https://gist.github.com/RavuAlHemio/262c054104688f81ea6d35395c2a1303)** — browser-console helper for the noVNC console; adjacent evidence of the same ecosystem of hacks against the VNC websocket.

Each of these reimplements a VNC client just to grab one picture, has to race the 1-minute token, and breaks if the proxy/RFB details change. A native endpoint would make all of that unnecessary.

## Proposed shape

Modeling on AWS/GCP, something like a read-style action:

```
POST /v1/servers/{id}/actions/request_console_screenshot
```

returning a small JSON body, e.g.:

```json
{
  "image": "<base64-encoded image data>",
  "content_type": "image/png"
}
```

(JPEG would be equally fine — matching AWS/GCP — as long as `content_type` makes it explicit.) No interactive session, no token-window race; the caller just decodes the base64 and writes a file.

CLI surface:

```
hcloud server screenshot <server> -o console.png   # decode + write the image
hcloud server screenshot <server> > console.png    # raw image to stdout
```

## Notes

- This is genuinely a platform capability, so I understand it may not be actionable in this repo directly — feel free to redirect me to the Console Feedback menu or to `hcloud-go`/the API team. I'd just like it tracked and the demand visible.
- Not urgent and not a blocker — the workarounds above function. The ask is to make the common "my VM won't boot, show me the screen" case a one-liner instead of a custom VNC client. Thanks for considering it.

jooola:
Adds a new `hcloud server screenshot -o screenshot.png <server>`, that take a screenshot of the server console using a VNC client.

Fixes #1420

### DIFF
--- a/docs/reference/manual/hcloud_server.md
+++ b/docs/reference/manual/hcloud_server.md
@@ -55,6 +55,7 @@ Manage Servers
 * [hcloud server request-console](hcloud_server_request-console.md)	 - Request a WebSocket VNC console for a Server
 * [hcloud server reset](hcloud_server_reset.md)	 - Reset a Server
 * [hcloud server reset-password](hcloud_server_reset-password.md)	 - Reset the root password of a Server
+* [hcloud server screenshot](hcloud_server_screenshot.md)	 - Take a screenshot of the Server VNC console
 * [hcloud server set-rdns](hcloud_server_set-rdns.md)	 - Change reverse DNS of a Server
 * [hcloud server shutdown](hcloud_server_shutdown.md)	 - Shutdown a server
 * [hcloud server ssh](hcloud_server_ssh.md)	 - Spawn an SSH connection for the Server
--- a/internal/cmd/server/screenshot/screenshot.go
+++ b/internal/cmd/server/screenshot/screenshot.go
@@ -1,0 +1,139 @@
+package screenshot
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"image"
+	"image/color"
+	"image/png"
+	"io"
+	"log/slog"
+	"net/url"
+	"os"
+
+	"github.com/alexsnet/go-vnc"
+	"github.com/alexsnet/go-vnc/rfbflags"
+	"golang.org/x/net/websocket"
+)
+
+func websocketOrigin(wsURL string) (string, error) {
+	u, err := url.Parse(wsURL)
+	if err != nil {
+		return "", err
+	}
+	u.Scheme = "https"
+	return u.String(), nil
+}
+
+func dialWebsocket(ctx context.Context, wsURL string) (*websocket.Conn, error) {
+	origin, err := websocketOrigin(wsURL)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg, err := websocket.NewConfig(wsURL, origin)
+	if err != nil {
+		return nil, err
+	}
+
+	ws, err := cfg.DialContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	ws.PayloadType = websocket.BinaryFrame
+
+	return ws, nil
+}
+
+func TakeScreenshot(ctx context.Context, wsURL string, filename string) error {
+	slog.Debug("dialing websocket")
+	ws, err := dialWebsocket(ctx, wsURL)
+	if err != nil {
+		return err
+	}
+	defer ws.Close()
+
+	slog.Debug("creating vnc client")
+	vncCfg := &vnc.ClientConfig{
+		Auth: []vnc.ClientAuth{
+			// Auth is managed at the HTTP/websocket level,therefore no password is
+			// needed for the VNC client.
+			&vnc.ClientAuthNone{},
+		},
+		ServerMessages: []vnc.ServerMessage{
+			&vnc.FramebufferUpdate{},
+		},
+		ServerMessageCh: make(chan vnc.ServerMessage, 1),
+	}
+
+	slog.Debug("connecting vnc client")
+	vncConn, err := vnc.Connect(ctx, ws, vncCfg)
+	if err != nil {
+		return err
+	}
+	defer vncConn.Close()
+
+	go func() {
+		slog.Debug("listening for vnc server messages")
+		if err := vncConn.ListenAndHandle(); err != nil {
+			slog.Error("failed to listen and handle incoming vnc server messages", "err", err)
+		}
+	}()
+
+	slog.Debug("requesting framebuffer update")
+	if err := vncConn.FramebufferUpdateRequest(
+		rfbflags.RFBFalse,
+		0, 0,
+		vncConn.FramebufferWidth(), vncConn.FramebufferHeight(),
+	); err != nil {
+		return err
+	}
+
+	slog.Debug("waiting for server message")
+	msg := <-vncCfg.ServerMessageCh
+
+	framebufferUpdate, ok := msg.(*vnc.FramebufferUpdate)
+	if !ok {
+		return fmt.Errorf("received unexpected server message: %T", msg)
+	}
+
+	// Sanity checks
+	if len(framebufferUpdate.Rects) != 1 ||
+		framebufferUpdate.Rects[0].X != 0 ||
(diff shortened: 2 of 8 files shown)


jooola:
Still missing from this PR:
- Finalize the e2e test.
- Replace slog logger with what the CLI is using.
- Hide log messages from the go-vnc library.
- Check if another fork og go-vnc is better that the one from alexnet, I saw other forks being a bit more active.
- Maybe `golang.org/x/net/websocket` is not the best library, but it does not add a new dependency. Check if this is important.

docs/reference/manual/hcloud_server.md
```diff
@@ -55,6 +55,7 @@ Manage Servers
 * [hcloud server request-console](hcloud_server_request-console.md)	 - Request a WebSocket VNC console for a Server
 * [hcloud server reset](hcloud_server_reset.md)	 - Reset a Server
 * [hcloud server reset-password](hcloud_server_reset-password.md)	 - Reset the root password of a Server
+* [hcloud server screenshot](hcloud_server_screenshot.md)	 - Take a screenshot of the Server VNC console
```
apricote, line 58: I would prefer to mark this new command as experimental initially.

apricote:
Would be awesome if users could directly show it in their terminal.

For example with the Kitty graphics protocol I can run this right now to render the image in the terminal (Ghostty):


```shell
hcloud server screenshot <name>
viu screenshot.png
```

If `-o -` would mean `stdout`:

```shell
hcloud server screenshot <name> -o - | viu -
```

internal/cmd/server/screenshot/screenshot.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	slog.Debug("connecting vnc client")
+	vncConn, err := vnc.Connect(ctx, ws, vncCfg)
+	if err != nil {
+		return err
+	}
+	defer vncConn.Close()
+
+	go func() {
```
phm07, line 77: The server message channel should be closed or we might end up waiting for it endlessly.

```suggestion
	go func() {
		defer close(vncCfg.ServerMessageCh)
```

internal/cmd/server/screenshot/screenshot.go
```diff
@@ -0,0 +1,139 @@
...
+	imgData := bytes.NewBuffer(nil)
+	{
+		pngEncoder := png.Encoder{CompressionLevel: png.DefaultCompression}
+		if err := pngEncoder.Encode(io.Writer(imgData), img); err != nil {
+			return err
+		}
+	}
+
+	slog.Debug("writing image to file")
+	if err := os.WriteFile(filename, imgData.Bytes(), 0600); err != nil {
```
phm07, line 134: This overwrites `screenshot.png` in the current dir without asking. Should we fail if the file already exists and add an explicit `--overwrite` flag?

PR #1445, opened 2026-07-02
feat: add http-timeout option

phm07:
This PR adds a configurable `http-timeout` option.
Since the default Go HTTP client has a timeout of 0, this is the proposed default for the new option as well to keep behavior consistent. A value of 0 means no timeout.


internal/state/state.go
```diff
@@ -135,6 +136,15 @@ func (c *state) newClient() (hcapi2.Client, error) {
 		}))
 	}
 
+	httpTimeout, err := config.OptionHTTPTimeout.Get(c.config)
+	if err != nil {
+		return nil, err
+	}
+
+	client := http.DefaultClient
```
jooola:
```suggestion
	client := &http.Client{}
```

Could we our own dedicated client, instead of the global default client?

phm07: Yes, makes sense

internal/state/config/options.go
```diff
@@ -169,6 +169,15 @@ var (
 		nil,
 	)
 
+	OptionHTTPTimeout = newOpt(
+		"http-timeout",
+		"Timeout for HTTP requests (0 = no timeout)",
```
jooola: This is not showing that 0 is the default value.

phm07: The other options do not have their defaults documented either. Do we want to put it in the description or maybe add an extra column in the help table?

jooola:
```suggestion
		"Timeout for HTTP requests (default 0 = no timeout)",
```

What about this?

jooola (approved):

PR #1449, opened 2026-07-06
feat: remove datacenter property from server and primary_ip

lukasmetzner:
Remove the `datacenter` property from Server and Primary IP resource where possible. The property was deprecated ~6 months ago, and because the property is removed from the API, removing it from our code is not considered a breaking change.


lukasmetzner:
Needs https://github.com/hetznercloud/hcloud-go/pull/890 and rebase

jooola:
Closed in favor of #1460

PR #1457, opened 2026-07-17
feat: add api command

### LINKED ISSUE #1448: Support generic `hcloud api` subcommand
### TL;DR

When accessing the `hcloud` API, especially for new products or internal APIs, I'd love to see some kind of `hcloud api /foo/bar` in order to reuse the parsing, authentication, etc. This would be similar to [`glab api`](https://docs.gitlab.com/cli/api/) or the `gh api` of the GitHub API.

### Expected behavior

I'd like to have the following behaviour defined:

`hcloud api /api/v1/fancy_new_product` should perform a GET request with the authentication token being set against `https://api.hetzner.cloud/api/v1/fancy_new_product`.

To perform other types of requests (POST, DELETE, PUT), I can imagine the following CLI:

```bash
$ hcloud api --field foo=bar -X PUT /api/v1/fancy_new_product
$ # equivalent to the following cURL
$ curl -X PUT https://api.hetzner.cloud/api/v1/fancy_new_product --json '{"foo": "bar"}' -H "Authentication: ..."
```

The behaviour could be defined as follows:

1. If a `--field` parameter is provided, we default to POST.
2. If no `--field` parameter is provided, it's GET.
3. `--field` parameters are added to a JSON object.

It might even make sense to add a `-1` parameter, which adds the `/api/v1` prefix automatically.

phm07:
This PR adds a command that allows users to directly interact with the API.

Closes #1448

### DIFF
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -78,11 +78,12 @@ var APICmd = base.Cmd{
 			return err
 		}
 
-		var respBody json.RawMessage
-		_, err = s.Client().Do(req, &respBody)
+		var respBuf bytes.Buffer
+		_, err = s.Client().Do(req, &respBuf)
 		if err != nil {
 			return err
 		}
+		respBody := respBuf.Bytes()
 
 		if json.Valid(respBody) {
 			var formatted bytes.Buffer
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	_ "embed"
 	"encoding/json"
 	"io"
 	"net/url"
@@ -15,8 +14,8 @@ import (
 	"github.com/hetznercloud/cli/internal/state"
 )
 
-var ApiCmd = base.Cmd{
-	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
+var APICmd = base.Cmd{
+	BaseCobraCommand: func(_ hcapi2.Client) *cobra.Command {
 		cmd := &cobra.Command{
 			Use:   "api [options] <path>",
 			Short: "Make API call",
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -1,0 +1,71 @@
+package api
+
+import (
+	_ "embed"
+	"encoding/json"
+	"io"
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/hetznercloud/cli/internal/cmd/base"
+	"github.com/hetznercloud/cli/internal/hcapi2"
+	"github.com/hetznercloud/cli/internal/state"
+)
+
+var ApiCmd = base.Cmd{
+	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
+		cmd := &cobra.Command{
+			Use:   "api [options] <path>",
+			Short: "Make API call",
+		}
+
+		cmd.Flags().StringP("method", "X", "GET", "HTTP method to use for API call")
+		cmd.Flags().StringP("data", "d", "", "HTTP request body content (use - to read from stdin)")
+		cmd.Flags().StringToStringP("value", "v", nil, "key=value pairs to pass as query parameters")
+		return cmd
+	},
+	Run: func(s state.State, cmd *cobra.Command, args []string) error {
+		path := args[0]
+		method, _ := cmd.Flags().GetString("method")
+		data, _ := cmd.Flags().GetString("data")
+		values, _ := cmd.Flags().GetStringToString("value")
+
+		var body io.Reader
+		switch data {
+		case "":
+			body = nil
+		case "-":
+			body = os.Stdin
+		default:
+			body = strings.NewReader(data)
+		}
+
+		u, err := url.Parse(path)
+		if err != nil {
+			return err
+		}
+
+		params := u.Query()
+		for k, v := range values {
+			params.Set(k, v)
+		}
+		u.RawQuery = params.Encode()
+
+		req, err := s.Client().NewRequest(s, method, u.String(), body)
+		if err != nil {
+			return err
+		}
+
+		var respBody json.RawMessage
+		_, err = s.Client().Do(req, &respBody)
+		if err != nil {
+			return err
+		}
+
+		cmd.Print(string(respBody))
+		return nil
+	},
+}
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -42,6 +42,10 @@ var APICmd = base.Cmd{
 			body = strings.NewReader(data)
 		}
 
+		if !strings.HasPrefix(path, "/") {
+			path = "/" + path
+		}
+
 		u, err := url.Parse(path)
 		if err != nil {
 			return err
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -3,7 +3,9 @@ package api
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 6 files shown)


internal/cmd/api/api.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/hetznercloud/cli/internal/cmd/base"
+	"github.com/hetznercloud/cli/internal/hcapi2"
+	"github.com/hetznercloud/cli/internal/state"
+)
+
+var ApiCmd = base.Cmd{
+	BaseCobraCommand: func(client hcapi2.Client) *cobra.Command {
+		cmd := &cobra.Command{
+			Use:   "api [options] <path>",
```
lukasmetzner, line 24: Not sure if this is relevant for the first iteration, but what about setting `Content-Type` and defaulting to json?

edit: we already set this in hcloud-go, so we would not need to default here.

internal/cmd/api/api.go
```diff
@@ -0,0 +1,71 @@
...
+			return err
+		}
+
+		var respBody json.RawMessage
+		_, err = s.Client().Do(req, &respBody)
+		if err != nil {
+			return err
+		}
+
+		cmd.Print(string(respBody))
```
lukasmetzner: Should we do pretty print with indents, or have it as an option?

phm07: You could pipe the output to `jq` for pretty printing. Not sure if we need to implement this functionality in the CLI

phm07:
We also need to think about how to handle switching between the `api.hetzner.cloud` and `api.hetzner.com` endpoints.

internal/cmd/api/api.go
apricote: I much prefer to pretty print in the CLI, as **I** always want the output to be prettified anyway.

nachtjasmin:
A minor user experience nitpick: Right now, the command requires me to prefix all paths with the forward slash `/`. It'd be nice if this could be added automatically to allow calls like `hcloud api fancy_product/123` instead of `hcloud api /fancy_product/123`. ^^

internal/cmd/api/api.go
apricote: Thanks :)

internal/cmd/api/api.go
```diff
@@ -0,0 +1,79 @@
...
+		_, err = s.Client().Do(req, &respBody)
+		if err != nil {
+			return err
+		}
+
+		var formatted bytes.Buffer
+		if err := json.Indent(&formatted, respBody, "", "  "); err != nil {
+			return err
+		}
+		cmd.Print(formatted.String())
```
apricote: This prints the formatted string without a final newline. In some shells this causes the prompt to be printed on the same line as the closing `}`. zsh shows a `%` after the closing `}` by default.

It looks like `jq` prints a final newline after formatted output. I would suggest we do the same, but maybe I am overlooking some drawbacks.

internal/cmd/api/api.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		cmd.Flags().StringP("method", "X", "GET", "HTTP method to use for API call")
+		cmd.Flags().StringP("data", "d", "", "HTTP request body content (use - to read from stdin)")
+		cmd.Flags().StringToStringP("value", "v", nil, "key=value pairs to pass as query parameters")
+		return cmd
+	},
+	Run: func(s state.State, cmd *cobra.Command, args []string) error {
+		path := args[0]
+		method, _ := cmd.Flags().GetString("method")
```
apricote, line 37: Using the `method` verbatim can cause issues if the user specifies a lower case method:

```shell
➜ hcloud api -X delete ssh_keys/115705306 --debug
--- Command:
"hcloud" "api" "-X" "delete" "ssh_keys/115705306" "--debug"



--- Request:
delete /v1/ssh_keys/115705306 HTTP/1.1
Host: api.hetzner.cloud
User-Agent: hcloud-cli/1.66.0-dev hcloud-go/2.44.0
Accept: application/json
Authorization: REDACTED
Accept-Encoding: gzip



--- Response:
HTTP/2.0 400 Bad Request
Content-Length: 150
Content-Type: text/html
Date: Fri, 24 Jul 2026 09:27:09 GMT

<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>nginx</center>
</body>
</html>


hcloud: hcloud: server responded with status code 400
```

I suggest we validate and autocomplete the valid options, and possibly `toUpper()` the input too.

internal/cmd/api/api.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		var respBody json.RawMessage
+		_, err = s.Client().Do(req, &respBody)
+		if err != nil {
+			return err
+		}
+
+		var formatted bytes.Buffer
+		if err := json.Indent(&formatted, respBody, "", "  "); err != nil {
```
apricote: This errors for empty responses (ie. DELETE with 204 status code):

```http
➜ hcloud api -X DELETE ssh_keys/115705306 --debug
--- Command:
"hcloud" "api" "-X" "DELETE" "ssh_keys/115705306" "--debug"



--- Request:
DELETE /v1/ssh_keys/115705306 HTTP/1.1
Host: api.hetzner.cloud
User-Agent: hcloud-cli/1.66.0-dev hcloud-go/2.44.0
Accept: application/json
Authorization: REDACTED
Accept-Encoding: gzip



--- Response:
HTTP/2.0 204 No Content
Access-Control-Allow-Headers: X-Requested-With,Authorization,Content-Type
Access-Control-Allow-Methods: OPTIONS,GET,POST,PUT,PATCH,DELETE
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Link,X-Correlation-ID
Access-Control-Max-Age: 86400
Date: Fri, 24 Jul 2026 09:27:19 GMT
Ratelimit-Limit: 5000
Ratelimit-Remaining: 4999
Ratelimit-Reset: 1784885239
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Correlation-Id: 9bb7811f61c56c27da34296d6ab5351d



hcloud: unexpected end of JSON input
```

internal/cmd/api/api.go
phm07: Should we autocomplete and validate all HTTP methods as per RFC 9110 Section 9.3 or only the CRUD ones that are actually used in the API? My suggestion would be that GET, POST, PUT, DELETE should be enough.

apricote: Autocomplete for the ones that are currently used and validate for all? That way suggestions are limited to a reasonable list, while it's still possible to use a new method if it's introduced in the API.

apricote: `PATCH` from [RFC 5789](https://www.rfc-editor.org/info/rfc5789/) may also be allowed in validation. Not sure why it is omitted in RFC 9110

internal/cmd/api/api.go
apricote: This still happens with [87cb8ff](https://github.com/hetznercloud/cli/pull/1457/commits/87cb8ff78f0b4f0ec45fe66e18f0e0e603b5fdaf). The error is returned from `s.Client().Do(req, &respBody)`, so we never get to the printing part in this file.

internal/cmd/api/api.go
phm07: I ended up validating against the methods defined [here](https://pkg.go.dev/net/http#pkg-constants)

internal/cmd/api/api.go
phm07: With d2e8e8b4cdac238750ddfe5cb72d7fc8392ca0d7 it should work

apricote: Confirmed, nice!

internal/cmd/api/api.go
apricote: Nice!

apricote (approved):
🚀

nachtjasmin:
it’s merged, yippie! thank you for your work, I really appreciate it! 💜

PR #1460, opened 2026-07-21
feat: remove datacenter property from server and primary_ip

lukasmetzner:
Remove the `datacenter` property from Server and Primary IP resource where
possible. The property was deprecated ~6 months ago, and because the property
is removed from the API, removing it from our code is not considered a breaking
change.

Changelog entry: https://docs.hetzner.cloud/changelog#2026-07-01-removing-datacenters


docs/reference/manual/hcloud_primary-ip_create.md
```diff
@@ -6,10 +6,7 @@ Create a Primary IP
 
 Create a Primary IP.
 
-The --datacenter flag is deprecated. Use --location or --assignee-id instead.
```
jooola, line 9: Could we maybe keep a little reference to the --datacenter in the docs? .e.g that is has been removed?

lukasmetzner: Added a little note to PrimaryIP and Server with a link to the changelog.

phm07 (approved):

internal/cmd/primaryip/create.go
```diff
@@ -21,11 +20,11 @@ var CreateCmd = base.CreateCmd[*hcloud.PrimaryIP]{
 			Short: "Create a Primary IP",
 			Long: `Create a Primary IP.
 
-The --datacenter flag is deprecated. Use --location or --assignee-id instead.
+The --datacenter flag has been removed. Use --location or --assignee-id instead.
 See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters.
 
-The --assignee-type flag will be required together with the --assignee-id flag. Using
```
phm07, line 27: Is this change on purpose or was this automatic formatting?

jooola (approved):

PR #1461, opened 2026-07-24
feat: print deprecated Load Balancer Type information

phm07:
This PR adds warnings for deprecated Load Balancer Types and displays the deprecation info in the `list` and `describe` commands.


phm07:
Waiting for next hcloud-go release so that we can use the new `deprecationutil`

jooola (approved):

jooola (changes requested):
- The lb type describe command does not print the deprecation info.
- The lb type list command does not have a "deprecated" column.
- The old lb type deprecated property is deprecation, the deprecation object must be used instead.

jooola:
Please merge https://github.com/hetznercloud/cli/pull/1459 and rebase this branch on top of it.

jooola (approved):

PR #1478, opened 2026-09-04
docs: add headers and review

svenja11:


lukasmetzner (approved):

docs/tutorials/setup-hcloud-cli.md
```diff
@@ -1,234 +1,325 @@
...
-If you see this output, congratulations! You have successfully set up the hcloud CLI on your local machine.
+- A [Hetzner Console account](https://console.hetzner.com).
+
+<br>
+
+----
+
+<br>
+
+1. Install the hcloud CLI
```
jooola, line 33: What is the reason for breaking the sections with titles into a numbered list and adding the many html elements?

svenja11: This is the formatting we use in the docs (see [Using S3 compatible CLI tools](https://docs.hetzner.com/storage/object-storage/getting-started/using-s3-api-tools), for example). I updated it like this so that the formatting is consistent. We usually use `<details>` when there are a lot of options, because it helps readers see all options at a glance, and makes the content easier to navigate in the docs.

jooola: Is there a way to find a middle ground between the format we had, and the one in the docs? Because now the tutorial does not look great on Github:

Before: https://github.com/hetznercloud/cli/blob/5b68ce0ae339987d6f42cd9a7cb9533be6354fb1/docs/tutorials/setup-hcloud-cli.md
After: https://github.com/hetznercloud/cli/blob/bd387d58bfdfeafb7003733f7e2dcac1913d9454/docs/tutorials/setup-hcloud-cli.md

In Github, having titles for the sections automatically generates a "Table of content" for the users:

Image

I feel like the markdown document is not the right place for fixing styling issues (e.g. adding `<br>` to have more space between paragraphs), could we maybe fix this in the docs code directly?

svenja11: I thought about it for a bit but I can't really think of a good in-between. If this formatting is causing issues on GitHub, I guess we can go back to the original formatting.

PR #1485, opened 2026-09-14
feat(image): print deprecated Image information

phm07:
See https://docs.hetzner.cloud/changelog#2026-09-08-removing-the-deprecated-field


internal/cmd/image/list.go
```diff
@@ -120,10 +120,10 @@ var ListCmd = &base.ListCmd[*hcloud.Image, schema.Image]{
 				return util.Age(image.Created, time.Now())
 			}).
 			AddFieldFn("deprecated", func(image *hcloud.Image) string {
-				if image.Deprecated.IsZero() {
+				if !image.IsDeprecated() {
 					return "-"
 				}
-				return util.Datetime(image.Deprecated)
+				return util.Datetime(image.UnavailableAfter())
```
jooola: The old value should be mapped to the `announced`,  not `unavailable_after`.

jooola (approved):

